Issue #89, opened 2022-01-31
Prometheus histogram matcher


sandreim:
Something like `alice: reports histogram pvf_execution_time has at last X samples in buckets ["0.2", "1.0", "3.0"]` IMO provides enough flexibility.

pepoviola:
Closed by  https://github.com/paritytech/zombienet/pull/94

PR #94, opened 2022-02-01
[Feat] add histogram support

pepoviola:
- Add histogram syntax

```
<node name>: reports histogram<metric name, with labels for match> has at least <x> samples in buckets <[array of buckets]> [within x seconds]
```

e.g: 
```
alice: reports histogram polkadot_pvf_execution_time has at least 2 samples in buckets ["0.1", "0.05", "0.025", "+Inf"] within 100 seconds
```

cc @sandreim

### DIFF
--- a/src/networkNode.ts
+++ b/src/networkNode.ts
@@ -1,7 +1,7 @@
 import { ApiPromise, WsProvider } from "@polkadot/api";
 import minimatch from "minimatch";
 
-import { Metrics, fetchMetrics, getMetricName } from "./metrics";
+import { Metrics, fetchMetrics, getMetricName, getHistogramBuckets, BucketHash } from "./metrics";
 import {
   DEFAULT_INDIVIDUAL_TEST_TIMEOUT,
   PROMETHEUS_PORT,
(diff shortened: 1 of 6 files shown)


src/networkNode.ts
```diff
@@ -254,6 +254,79 @@ export class NetworkNode implements NetworkNodeInterface {
...
+        clearTimeout(limitTimeout);
+        return value;
+      }
+
+      // loop until get the desired value or timeout
+      let done = false;
+      let c = 0;
+      while (!done) {
+        if (expired) throw new Error(`Timeout(${timeout}s)`);
+        await new Promise((resolve) => setTimeout(resolve, 1000));
```
sandreim, line 286: We should have a constant for this.

src/networkNode.ts
```diff
@@ -254,6 +254,79 @@ export class NetworkNode implements NetworkNodeInterface {
...
+      if (desiredMetricValue === null || value >= desiredMetricValue) {
+        debug(`value: ${value} ~ desiredMetricValue: ${desiredMetricValue}`);
+        clearTimeout(limitTimeout);
+        return value;
+      }
+
+      // loop until get the desired value or timeout
+      let done = false;
+      let c = 0;
+      while (!done) {
```
sandreim, line 284: Would this scale well when we have grouping and 100+ nodes ?

tests/0010-histogram-smoke-test.toml
```diff
@@ -0,0 +1,33 @@
+[settings]
+timeout = 1000
+bootnode = true
+# provider = "Podman"
+
+[relaychain]
+default_image = "{{ZOMBIENET_INTEGRATION_TEST_IMAGE}}"
+chain = "rococo-local"
+#default_command = "../polkadot/target/release/polkadot"
```
sandreim, line 9: This is currently a pain to switch between CI and local tests

sandreim (approved):

src/networkNode.ts
pepoviola: Yes, sounds good.

src/networkNode.ts
pepoviola: We will need to change some of the logic for `group` testing, is part of the refactoring that I'm working.
Thx!

pepoviola:
closes https://github.com/paritytech/zombienet/issues/89

PR #115, opened 2022-02-15
github action for releases added

emamihe:
With this commit the repository will be able to generate the binaries both for linux and macos automatically through github actions and push them to releases section by tag versions starting with v*

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,51 @@
+name: release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+jobs:
+  build_release:
+    name: build_release
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -0,0 +1,51 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: setup node
+        uses: actions/setup-node@v2
+        with:
+          node-version: '12'
```
pepoviola, line 20:
```suggestion
          node-version: '14'
```
At some point we need to change to `16` also.

pepoviola:
Thanks @emamihe!!

pepoviola (approved):

PR #167, opened 2022-04-12
Grep for listen connections for collators in native provider

pepoviola:
- Fix, `ready` state for collator in native provider.


emamihe (approved):
looks okay

PR #173, opened 2022-04-28
fix: proper variables overriding

sergejparity:
This PR fixed CI variables overriding. As doing so inside `variables` section does not work properly.


pepoviola (approved):

pepoviola:
Thanks @sergejparity!!

PR #198, opened 2022-05-12
Fix wait message to check if node is up

skunert:
I am using the `native` provider locally and was wondering why node start is not being detected. Turns out the logs do not print `Listening for new connections` anymore as of recently.


pepoviola:
Hi @skunert, yes I get the same issue today (I think is because polkadot now is using `jsonrpsee`).

Thanks for fixing 👍

PR #215, opened 2022-06-05
`0001-simple-network` to `0001-small-network`

shawntabrizi:
Seems like the example name changed, but examples are still using the old name.


pepoviola (approved):

pepoviola:
Hi @shawntabrizi, yes that was the case. 
Thanks!

PR #291, opened 2022-08-17
Fix setting the provider via file

atomaka:
By setting a default for the `--provider` cli argument and allowing the `--provider` cli argument to override the config setting if present, it is impossible to set the provider via config since a value will always be available for the `--provider` cli argument.

This chain breaks a nested if statement into two specific pieces:
1. If the `--provider` cli argument is provided, override the config setting
2. If config is missing for provider or timeout, use the default values.

## Notes

This changes the condition when `DEFAULT_GLOBAL_TIMEOUT` is set. Previously, it was only set when `--provider` was provided and `config.settings` was empty. With this change, `DEFAULT_GLOBAL_TIMEOUT` will be set whenever `config.settings.timeout` is not set.

### DIFF
--- a/src/cli.ts
+++ b/src/cli.ts
@@ -8,7 +8,11 @@ import { getCredsFilePath, readNetworkConfig } from "./utils/fs-utils";
 import { LaunchConfig } from "./types";
 import { run } from "./test-runner";
 import { Command, Option } from "commander";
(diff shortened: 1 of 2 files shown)


pepoviola (approved):

pepoviola:
Hi @atomaka, Looks good! Merging now, thanks!!

PR #304, opened 2022-08-22
Rename mergify file

wirednkod:
Fix erroneously named `.mergify.yml` to `mergify.yml`.
(That is why it was not working before)


pepoviola (approved):

pepoviola:
Thanks @wirednkod!!

PR #312, opened 2022-08-25
Create Setup process

wirednkod:
This PR creates an automated script that, setups locally the environment for running zombienet. Image below shows the outcome of the script:

image

User, once finishes with cloning and building the repo, can run the command `node dist/setup.js -b <binaries>`:
This will add the binaries that user wants to be downloaded, provided in a row without any separators; They are downloaded in 'dist' directory and appropriate executable permissions are assigned. 
Possible options: `polkadot`, `parachain`. Example: `node dist/setup.js -b polkadot parachain`

`AdderCollator` also is a possible option but due to the fact that it exists only in gitlab as an artifact that seizes to exist after few days, it is not activated just yet

### DIFF
--- a/src/constants.ts
+++ b/src/constants.ts
@@ -11,14 +11,6 @@ export const RPC_HTTP_PORT = 9944;
 // The port substrate listens for p2p connections on
 export const P2P_PORT = 30333;
 
-// Paths for downloading the binaries for zombienet setup
-export const DEFAULT_COMMAND_URL =
-  "https://github.com/paritytech/polkadot/releases/download/v0.9.27/polkadot";
-export const DEFAULT_ADDER_COLLATOR_URL =
-  "https://gitlab.parity.io/parity/mirrors/polkadot/-/jobs/1769497/artifacts/raw/artifacts/adder-collator";
-export const DEFAULT_CUMULUS_COLLATOR_URL =
-  "https://github.com/paritytech/cumulus/releases/download/v0.9.230/polkadot-parachain";
-
 export const DEFAULT_GLOBAL_TIMEOUT = 1200; // 20 mins
 export const DEFAULT_INDIVIDUAL_TEST_TIMEOUT = 10; // seconds
 export const DEFAULT_COMMAND = "polkadot";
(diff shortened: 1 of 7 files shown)


src/constants.ts
```diff
@@ -11,6 +11,14 @@ export const RPC_HTTP_PORT = 9944;
 // The port substrate listens for p2p connections on
 export const P2P_PORT = 30333;
 
+// Paths for downloading the binaries for zombienet setup
+export const DEFAULT_COMMAND_URL =
+  "https://github.com/paritytech/polkadot/releases/download/v0.9.27/polkadot";
```
pepoviola, line 16: Hi @wirednkod, I think we need to add the logic to download the latest release so we don't need to update this path every new release. For example `v0.9.28` was release two days ago.

src/constants.ts
```diff
@@ -11,6 +11,14 @@ export const RPC_HTTP_PORT = 9944;
...
+export const DEFAULT_ADDER_COLLATOR_URL =
+  "https://gitlab.parity.io/parity/mirrors/polkadot/-/jobs/1769497/artifacts/raw/artifacts/adder-collator";
+export const DEFAULT_CUMULUS_COLLATOR_URL =
+  "https://github.com/paritytech/cumulus/releases/download/v0.9.230/polkadot-parachain";
```
pepoviola, line 20: same here.

pepoviola:
Hi @wirednkod, looks good!!! thanks! The only comment is try to get the latest version querying the github api if is possible.

Thanks!

src/constants.ts
wirednkod: Could not agree more. This was more of a draft version, to see if we are on correct path

src/constants.ts
wirednkod: For Parachain URL we cannot really do that as the latest release may not contain the `polkadot-parachain` asset (see [the latest one](https://github.com/paritytech/cumulus/releases/tag/parachains-v9270) for example. I will make a change for the polkadot, but we will need probably, until somehow this is standardized, to adjust the `polkadot-parachain` manually for now

pepoviola (approved):
Awesome!! 🙌🙌 Thanks!!

PR #320, opened 2022-08-26
Upgrade `execa` correctly

wirednkod:
In order to upgrade `execa` from [v5 -> v6](https://github.com/paritytech/zombienet/pull/295) some changes in code must take place. Read more about this upgrade [here](https://github.com/sindresorhus/execa/releases/tag/v6.0.0)


wirednkod:
We cannot actually do that as this way we will not support anymore the older versions (CJS). It seems that `execa` [moved to esm](https://gist.github.com/sindresorhus/a39789f98801d908bbc7ff3ecc99d99c), so I am closing this.

PR #331, opened 2022-08-29
[poc] collect logs as artifacts

pepoviola:


wirednkod:
Is there a reason why we should "break" the dir and bring the logs as artifacts? I mean - was it requested from a team that uses zombienet?

pepoviola:
> Is there a reason why we should "break" the dir and bring the logs as artifacts? I mean - was it requested from a team that uses zombienet?

Yes, we want to give access to logs for externar contributors (e.g for polkadot/cumulus repos) that doesn't have access to our internal infra. 

Thanks!

pepoviola:
Closing here since is already implemented.

PR #354, opened 2022-09-02
Fix rest of printouts on CLI

wirednkod:


src/providers/k8s/kubeClient.ts
```diff
@@ -330,7 +333,6 @@ export class KubeClient extends Client {
     if (unique) {
       const args = ["cp", localFilePath, `${identifier}:${podFilePath}`];
       if (container) args.push("-c", container);
-      const result = await this.runCommand(args, undefined, true);
```
pepoviola, line 333: Why we are removing this line? we should do
```js
await this.runCommand(args, undefined, true);
```

wirednkod: Erroneously probably during some merge. Fixed it. Apologies about that

pepoviola: Awesome, don't need to apology :)

pepoviola (approved):

PR #377, opened 2022-09-12
changes for support node-template

pepoviola:
Add support for running a network based on [node-template](https://github.com/substrate-developer-hub/substrate-node-template)

### DIFF
--- a/src/chain-spec.ts
+++ b/src/chain-spec.ts
@@ -225,7 +225,7 @@ export async function addGrandpaAuthority(
 
   const chainSpec = readAndParseChainSpec(specPath);
 
-  let keys = getAuthorityKeys(chainSpec, "grandpa");
+  const keys = getAuthorityKeys(chainSpec, "grandpa");
   if (!keys) return;
 
   keys.push([ed_account.address, 1]);
--- a/src/chain-spec.ts
+++ b/src/chain-spec.ts
@@ -10,10 +10,10 @@ const debug = require("debug")("zombie::chain-spec");
 // track 1st staking as default;
 let stakingBond: number | undefined;
 
-export type KeyType = "session" | "aura";
+export type KeyType = "session" | "aura" | "grandpa";
 
 // Check if the chainSpec have session keys
-export function specHaveSessionsKeys(chainSpec: ChainSpec) {
+export function specHaveSessionsKeys(chainSpec: ChainSpec): boolean {
   // Check runtime_genesis_config key for rococo compatibility.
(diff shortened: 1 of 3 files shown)


src/chain-spec.ts
```diff
@@ -210,6 +217,28 @@ export async function addAuraAuthority(
   );
 }
 
+export async function addGrandpaAuthority(
+  specPath: string,
+  name: string,
+  accounts: any,
```
wirednkod, line 223: IS it possible to avoid `any` and define (create) a type for accounts? It would save some errors

src/chain-spec.ts
```diff
@@ -210,6 +217,28 @@ export async function addAuraAuthority(
...
+) {
+  const { ed_account } = accounts;
+
+  const chainSpec = readAndParseChainSpec(specPath);
+
+  let keys = getAuthorityKeys(chainSpec, "grandpa");
```
wirednkod, line 229:
```suggestion
  const keys = getAuthorityKeys(chainSpec, "grandpa");
```

src/chain-spec.ts
```diff
@@ -210,6 +217,28 @@ export async function addAuraAuthority(
...
+  if (!keys) return;
+
+  keys.push([ed_account.address, 1]);
+
+  writeChainSpec(specPath, chainSpec);
+  console.log(
+    `\t👤 Added Genesis Authority (GRANDPA) ${decorators.green(
+      name,
+    )} - ${decorators.magenta(ed_account.address)}`,
+  );
```
wirednkod, line 239: Maybe use the [logtable](https://github.com/paritytech/zombienet/blob/main/src/utils/logger.ts#L56) instead of the console log?

wirednkod (approved):
Minor things - other than those - it looks good to me

src/chain-spec.ts
pepoviola: I'm not sure if we want to use `logtable` for all logs, but we can chat about it 👍

wirednkod: Not a priortiy 1 at the moment. So lets not worry about it

src/chain-spec.ts
pepoviola: Agree, we should try to avoid the use of `any` as much as possible.
I will open a separate issue to track this, thanks!

pepoviola:
@wirednkod, I will merge this one an open a new issue related to the refactor of the usage of `any`.
Thanks!

PR #387, opened 2022-09-14
fix args for run in k8s

pepoviola:
@wirednkod small fix.


pepoviola:
@wirednkod, merging since is a very small fix.

PR #391, opened 2022-09-16
fix, restart port-fw on node restart

pepoviola:
Recreate the port-fw when a node is restarted.

### DIFF
--- a/src/networkNode.ts
+++ b/src/networkNode.ts
@@ -200,9 +200,9 @@ export class NetworkNode implements NetworkNodeInterface {
         return;
       };
 
-      // force connect
-      if (this.apiInstance) (this.apiInstance as ApiPromise).disconnect();
-      await this.connectApi();
+      // // force connect
+      // if (this.apiInstance) (this.apiInstance as ApiPromise).disconnect();
+      // await this.connectApi();
 
       const resp = await Promise.race([
         getValue(),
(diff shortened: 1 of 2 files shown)


src/networkNode.ts
```diff
@@ -77,7 +83,26 @@ export class NetworkNode implements NetworkNodeInterface {
...
+          "{{PORT}}",
+          fwdPort.toString(),
+        );
+
+        this.apiInstance = undefined;
+      }
+
+      return true;
+    }
+    return false;
```
wirednkod, line 105: Its better to do an early exit
```suggestion
    if(result.exitCode !== 0) return false;
    // restart the port-fw if needed
    const url = new URL(this.wsUri);
    if (parseInt(url.port, 10) !== RPC_WS_PORT) {
      const fwdPort = await client.startPortForwarding(
        RPC_WS_PORT,
        this.name,
      );

    this.wsUri = WS_URI_PATTERN.replace("{{IP}}", LOCALHOST).replace(
      "{{PORT}}",
      fwdPort.toString(),
    );

    this.apiInstance = undefined;
    return true;
```

src/networkNode.ts
```diff
@@ -179,6 +204,10 @@ export class NetworkNode implements NetworkNodeInterface {
         return;
       };
 
+      // // force connect
+      // if (this.apiInstance) (this.apiInstance as ApiPromise).disconnect();
+      // await this.connectApi();
```
wirednkod, line 209: Please remove commented code (If its a TODO thing, just add that to a comment )

src/networkNode.ts
pepoviola: Yes, sounds better 👍

src/networkNode.ts
pepoviola: Ahh nice catch, I will remove those lines. Thanks!

wirednkod (approved):

PR #411, opened 2022-09-21
Convert Polkadot Launch config to Zombienet

wirednkod:
Convert a given config file from project `previous-launch` to zombienet config `.toml` and `.json`. Files are placed in same dir as the given file with the addition of `-zombie` to the filename.

-Missing (and ofc TODOs): Add support convertion for genesis customization / hrmpChannels and types


src/types.d.ts
```diff
@@ -255,3 +255,53 @@ export interface Resources {
...
+            validation_upgrade_frequency: number;
+            validation_upgrade_delay: number;
+          };
+        };
+      };
+    };
+  };
+}
+
+interface ParaChainConfig {
```
pepoviola, line 285: We have a similar intergace `ParachainConfig` we should use a prefix like `PL_` or maybe move those to a diff types.d file.

src/types.d.ts
```diff
@@ -255,3 +255,53 @@ export interface Resources {
...
+  bin?: string;
+  id?: string;
+  port?: string;
+  balance?: string;
+  nodes: [NodesConfig];
+}
+
+interface HrmpChannels {
```
pepoviola, line 293: same here.

src/cli.ts
```diff
@@ -119,6 +122,117 @@ const latestPolkadotReleaseURL = async (
...
+async function convertInput(filePath: string) {
+  const { fullPath, fileName, extension } = getFilePathNameExt(filePath);
+
+  const convertedJson = await readInputFile(extension, filePath);
+
+  stream = startStream(`${fullPath}/${fileName}-zombie.toml`);
+
+  const { relaychain, parachains, simpleParachains } = convertedJson;
+
+  if (relaychain) {
```
pepoviola, line 188: Hey @wirednkod, we can also simplify the conversion by work with `json` directly in js and the use something like [json2toml](https://www.npmjs.com/package/json2toml) to output `toml`. What do you think?

wirednkod: 🤔 Looks more decent than the solution below. Thank you @pepoviola - will make the convertion

pepoviola (approved):
Looks good!! thanks!

PR #414, opened 2022-09-22
Parachain chain_spec bug fix

### LINKED ISSUE #409: `chain_spec_path` property for parachains is ignored
I assumed that when you provide `chain_spec_path` to parachain configuration, it should override manual chain spec generation and just copy provided file to the temp folder. But `parachains.chain_spec_path` argument's value is not passed to chain spec generation method `setupChainSpec()` [here](https://github.com/paritytech/zombienet/blob/main/src/paras.ts#L53). This makes `chain_spec_path` argument for parachain configuration useless.

dastanbeksamatov:
### Chain spec generation

This PR fixes #409 by giving `chain_spec_path` higher priority than `chain_spec_command` for chain specification generation. I think it makes more sense to check first if path to the chain spec is provided and only generate the chain spec manually when needed.


dastanbeksamatov:
hey @pepoviola,

thanks for the quick response. It's great news that session keys and chain spec generation will be customisable! can't wait for it!

here is `SessionKeys` struct of `efinity` and `rocfinity` (`rococo` parachain):

```rs
pub struct SessionKeys {
    pub aura: Aura,
    pub pools: Pools,
}
```

I will also remove code that affects session keys from this PR and will just leave chain spec part.

pepoviola (approved):

pepoviola:
thanks @dastanbeksamatov!!

PR #415, opened 2022-09-23
Feat decorators

pepoviola:
Add _decorators_ logic to allow chain customization by parachains. Each parachain can add his customization by overriding one of the `fns` that customize the chain-spec

```
    addAuraAuthority,
    addAuthority,
    changeGenesisConfig,
    clearAuthorities,
    readAndParseChainSpec,
    specHaveSessionsKeys,
    getNodeKey,
    addParaCustom,
    addCollatorSelection,
    writeChainSpec,
 ```

by adding the parachain in the `enum` (in `para-decorators/index.ts`)
 
```
enum PARA {
    Statemint = "statemint",
    Moonbeam = "moonbeam",
    Generic = "generic"
   }
```

Then import the file with the overrides _fns_
```
// imports
import moonbeam from "./moonbeam";
import statemint from "./statemint";
```

And  finally add to the registry

```
const statemintDecorators: ParaDecorator = Object.keys(statemint).reduce((memo, fn) => {
    memo[fn] = (statemint as ParaDecorator)[fn];
    return memo;
},Object.create({}));

const decorators : {[para in PARA]: { [fn: string]: Function}} = {
    moonbeam : moonbeamDecorators,
    statemint: statemintDecorators,
    generic: {}
};
```

cc @crystalin, I moved the moonbeam specific code and added a smoke test to ensure block production.

PR #422, opened 2022-09-26
Add efinity decorators and fix a bug

dastanbeksamatov:
- Add `efinity` parachain decorators
- Fix regex matching bug.  

  `(chain.match(/["moonbase"|"moonriver"|"moonbeam"]/gi))` will be true most of the times, e.g whenever at least one letter of `chain`'s value matches any letter of `moonbeam|moonriver|moonbase`. This is why it was detecting `efinity` as `moonbeam` chain when I was passing `efinity-local` as `parachain.chain` value.

P.S Good job on #415! Was looking forward for it and am looking forward for using `zombienet` with Efinity :)


dastanbeksamatov:
@pepoviola looking forward for your review

pepoviola (approved):
Looks great! thanks for your contribution and for fix the regex :)

PR #426, opened 2022-09-27
add logic to generate random nominators

pepoviola:
Add new `relaychain` config `random_nominators_count`, to generate random nominators and add those into the chain-spec genesis.

cc @ordian

### DIFF
--- a/src/chain-spec.ts
+++ b/src/chain-spec.ts
@@ -3,7 +3,9 @@ import { decorators } from "./utils/colors";
 import { ChainSpec, HrmpChannelsConfig, Node } from "./types";
 import { readDataFile } from "./utils/fs";
 import { convertExponentials } from "./utils/misc";
+import { generateKeyFromSeed } from "./keys";
 import fs from "fs";
+import crypto from 'crypto';
 const JSONbig = require("json-bigint")({ useNativeBigInt: true });
 const debug = require("debug")("zombie::chain-spec");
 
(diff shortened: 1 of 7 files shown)


src/chain-spec.ts
```diff
@@ -433,6 +467,19 @@ export function writeChainSpec(specPath: string, chainSpec: any) {
   }
 }
 
+// helper
+function getRandom(arr: string[], n: number) {
```
wirednkod, line 475: I think it is better for helpers to exist under `utils`?

src/configGenerator.ts
```diff
@@ -72,6 +72,7 @@ export async function generateNetworkSpec(
       defaultImage: config.relaychain.default_image || DEFAULT_IMAGE,
       defaultCommand: config.relaychain.default_command || DEFAULT_COMMAND,
       defaultArgs: config.relaychain.default_args || [],
+      randomNominatorsCount: config.relaychain.random_nominators_count || 0,
```
wirednkod, line 75:
```suggestion
      randomNominatorsCount: config.relaychain?.random_nominators_count || 0,
```

src/keys.ts
```diff
@@ -9,6 +9,14 @@ function nameCase(string: string) {
   return string.charAt(0).toUpperCase() + string.slice(1);
 }
 
+export async function generateKeyFromSeed(seed: string): Promise<any> {
+  await cryptoWaitReady();
+
+  const sr_keyring = new Keyring({ type: "sr25519" });
+  const sr_account = sr_keyring.createFromUri(`//${seed}`);
+  return sr_account;
```
wirednkod, line 17:
```suggestion
  return sr_keyring.createFromUri(`//${seed}`);
```

src/chain-spec.ts
pepoviola: Yes, we can move to `utils` since is very generic.

pepoviola: done, thanks!!

wirednkod (approved):

PR #440, opened 2022-09-30
Update README.md

tobbelobb:
The  `--provider native` option was missing in a usage example

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,7 +174,7 @@ id = 100
 Then you can spawn the network by running the following command:
 
 ```bash
-./zombienet-macos spawn examples/0001-small-network.toml --provider native
+./zombienet-macos spawn --provider native examples/0001-small-network.toml
 ```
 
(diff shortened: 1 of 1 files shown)


wirednkod:
> The  `--provider native` option was missing in a usage example

It could run without it (defaults to k8s) but it makes sense since the outcome in the example notes that the provider is the native. Thank you

README.md
```diff
@@ -174,7 +174,7 @@ id = 100
 Then you can spawn the network by running the following command:
 
 ```bash
-./zombienet-macos spawn examples/0001-small-network.toml
+./zombienet-macos spawn examples/0001-small-network.toml --provider native
```
wirednkod, line 177:
```suggestion
./zombienet-macos spawn --provider native examples/0001-small-network.toml
```

wirednkod: The --provider arg should be before the filename

wirednkod (changes requested):

wirednkod (approved):

PR #441, opened 2022-10-01
Build instructions for adder-collator

tobbelobb:
I noticed that you previously used a binary download from Parity's internal gitlab jobs, but it's currently commented out in `dist/cli.js`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,6 +177,8 @@ Then you can spawn the network by running the following command:
 ./zombienet-macos spawn --provider native examples/0001-small-network.toml
 ```
 
+Note that the command expects two binaries `polkadot` and `adder-collator` to be installed on your system. See further down for how to get them.
+
 You can follow the output of the `steps` to spawn the network and once the network is launched a
 message with the `node`s information like this one is shown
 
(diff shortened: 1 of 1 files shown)


wirednkod:
Thank you for your PR @tobbelobb . The `adder-colllator` is meant to be removed. To explain this: When setup was written it was in plan to be used, but eventually it was commented out until we figure out if it will be used or not, and how exactly (as the "package" is built only in Gitlab CI for now). Eventually we decided that it will not be used and all examples will be altered to use only `polkadot` and `polkadot-parachain` only.

tobbelobb:
Ah I see. I suggest accepting this and reverting this at the same time as examples are updated.

wirednkod:
> Ah I see. I suggest accepting this and reverting it again in the commit that updates the examples.

You have a point. I will reopen

wirednkod (approved):

pepoviola:
Yes, I think we should include this until we move all the examples to use the `polkadot-parachain` from `cumulus`.
Thanks @tobbelobb!!

wirednkod:
Awesome. Thank you @tobbelobb

README.md
```diff
@@ -342,6 +344,18 @@ At the end of the download, the `setup` script will provide a command to run in
...
 
+### Build adder-collator (needed for running examples with native provider)
+
+You can build it from source like this
+
+```bash
+git clone git@github.com:paritytech/polkadot
+cd polkadot/parachain/test-parachains/adder/collator/
+cargo build
+mv ../../../../target/debug/adder-collator /home/<user>/zombienet/dist
```
pepoviola, line 355:
```suggestion
cd polkadot
cargo build --profile testnet --verbose -p test-parachain-adder-collator
export PATH=$(pwd)/target/testnet:$PATH
```

pepoviola:
Hi @tobbelobb,  nice catch! and thanks for your contribution. I make an small suggestion on the bash cmd.
Thanks!!

PR #450, opened 2022-10-04
[Fix] add logic to use max_nominations and fix raw spec creation in k8s

pepoviola:
- Fix nominations logic to allow set `max_nominations`.
- Fix chain spec creation (raw) in k8s.
- Add docs for new keys.


wirednkod (approved):
Lgtm

PR #472, opened 2022-10-12
Fix parser for custom-js/custom-sh

pepoviola:


pepoviola:
@wirednkod I will merge this one since the parser had a bug for `custom-js` with args.
Thanks!

PR #500, opened 2022-10-20
fix release artifacts

pepoviola:
Fix release ci

- Fix path to binaries, this path is relative to the process dir.
- Fix typo in macos version
- Fix working-dir for npm and use npm update for optional mac deps.


wirednkod (approved):
lgtm

PR #524, opened 2022-10-30
Add dir and force options in CLI

### LINKED ISSUE #264: [feat] Add `logs` attribute
Add `logs` attribute to `relaychain.nodes` and `parachains.collators` to be able to override the default path where logs are saved.

wirednkod:
This PR addresses the need of having specific directory for running zombienet instead of a random/tmp one.

Can be reviewed by each commit.

What is added:
- New option of `--dir`, `-d` : A specific directory path can be passed to the CLI; Zombienet files/logs etc will be all under that CLI. If directory does not exist, it will be created (and all files under it); If directory exists, then user will be prompted for his wish to overwrite it or not;
-  New option of `--force`, `-f`:  This options overwrites any user prompts that may occur during the CLI run (e.g. the "Directory already exists - would you like to overwrite it")
- Add checks for all the directories - if they exist do not create them 
- Update documentation and README files

Fixes #264

### DIFF
--- a/README.md
+++ b/README.md
@@ -126,6 +126,8 @@ Options:
   -c, --spawn-concurrency <concurrency>  Number of concurrent spawning process to launch, default is 1
   -p, --provider <provider>              Override provider to use (choices: "podman", "kubernetes", "native")
                                          default: kubernetes
+  -d, --dir <path>                       Directory path for placing the network files instead of random temp one
+  -f, --force                            Force override all prompt commands
   -m, --monitor                          Start as monitor, do not auto cleanup network
   -h, --help                             display help for command
 
--- a/README.md
+++ b/README.md
@@ -126,7 +126,7 @@ Options:
   -c, --spawn-concurrency <concurrency>  Number of concurrent spawning process to launch, default is 1
   -p, --provider <provider>              Override provider to use (choices: "podman", "kubernetes", "native")
                                          default: kubernetes
-  -d, --dir <path>                       Directory path for placing the network files instead of random temp one
+  -d, --dir <path>                       Directory path for placing the network files instead of random temp one (e.g. -d /home/user/my-zombienet)
   -f, --force                            Force override all prompt commands
   -m, --monitor                          Start as monitor, do not auto cleanup network
   -h, --help                             display help for command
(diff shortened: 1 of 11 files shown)


javascript/packages/orchestrator/src/orchestrator.ts
```diff
@@ -117,8 +121,26 @@ export async function start(
...
 
-    // create tmp directory to store needed files
-    const tmpDir = await tmp.dir({ prefix: `${namespace}_` });
+    // use provided dir (and make some validations) or create tmp directory to store needed files
+    const tmpDir = opts.dir
+      ? { path: opts.dir }
+      : await tmp.dir({ prefix: `${namespace}_` });
+
+    // If custom path is provided then create it
+    if (opts.dir && !fs.existsSync(opts.dir)) {
```
pepoviola, line 130: this should be something like this...

```
if(!opts.dir || !fs.existsSync(opts.dir)) {
// create if we don't have the the `dir` set or is not exist
} else {
  // dir exist
  if(!force) ask
}
```

README.md
```diff
@@ -126,6 +126,8 @@ Options:
   -c, --spawn-concurrency <concurrency>  Number of concurrent spawning process to launch, default is 1
   -p, --provider <provider>              Override provider to use (choices: "podman", "kubernetes", "native")
                                          default: kubernetes
+  -d, --dir <path>                       Directory path for placing the network files instead of random temp one
```
pepoviola, line 129:
```suggestion
  -d, --dir <path>                       Directory path for placing the network files instead of random temp one (e.g. -d /home/user/my-zombienet)
```

javascript/packages/orchestrator/src/orchestrator.ts
```diff
@@ -117,8 +121,26 @@ export async function start(
...
+      fs.mkdirSync(opts.dir);
```
pepoviola, line 131:
```suggestion
      fs.mkdirSync(tmpDir.path);
```

pepoviola (approved):
We should also make a follow up pr to add a test in our CI using the dir option.

PR #552, opened 2022-11-01
remove ci check for versions; and update monorepo version to 0.1.0

wirednkod:


javascript/package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "@parity/zombienet",
-  "version": "1.3.10",
+  "name": "zombienet",
+  "version": "0.1.0",
```
pepoviola, line 3: why we are changing this?

pepoviola (approved):

javascript/package.json
wirednkod: No need to keep the same name and version as CLI. Its the root of repo and we are going to update just that.

PR #563, opened 2022-11-03
Fix release gha

pepoviola:
- Fix path for upload artifacts in release.
- Bump version to draft a new release.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -46,7 +46,7 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./bins/zombienet-linux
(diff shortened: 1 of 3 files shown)


.github/workflows/release.yaml
```diff
@@ -46,7 +46,7 @@ jobs:
           GITHUB_TOKEN: ${{ github.token }}
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
-          asset_path: ./bins/zombienet-linux
+          asset_path: ./javascript/packages/bins/zombienet-linux
```
wirednkod, line 49: packages directory is meant for the workspaces but not for the bins. I would suggest to use a bin directory under javascript

pepoviola: Oks, I will change the output directory of `pkg`.

wirednkod (approved):

PR #571, opened 2022-11-08
add registry for `podman` images

pepoviola:
- Add registry for `podman` (infra) images.

cc @drahnr


drahnr (approved):
:heart: thank you!

PR #582, opened 2022-11-11
Check if tracing is enabled

pepoviola:
For `podman` we assume that the tracing (jaeger) is enabled in the node and that wrong, for example `substrate` don't use it. So, now you can set `enable_tracing` to false and will work with `substrate`.

cc @michalkucharczyk

### DIFF
--- a/javascript/packages/orchestrator/src/orchestrator.ts
+++ b/javascript/packages/orchestrator/src/orchestrator.ts
@@ -407,7 +407,10 @@ export async function start(
 
     const monitorIsAvailable = await client.isPodMonitorAvailable();
     let jaegerUrl: string;
-    if (client.providerName === "podman" && networkSpec.settings.enable_tracing) {
+    if (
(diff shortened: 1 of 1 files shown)


michalkucharczyk:
Should I use it like this?
```
[relaychain]
tracing_enabled = false
```

pepoviola:
> Should I use it like this?
> 
> ```
> [settings]
> enable_tracing = false
> ```

Yes, exactly.
Thanks!

michalkucharczyk (approved):

michalkucharczyk:
Maybe the option could be renamed to `enable_jaeger_tracing` ?

pepoviola:
> Maybe the option could be renamed to `enable_jaeger_tracing` ?

Yes, make sense. We are planning to start working in the `v2` of zombienet. I think this rename make sense there and keep the `v1` as is now.

wirednkod (approved):

Issue #598, opened 2022-11-16
Add expression allowing to check count of log lines matching pattern


michalkucharczyk:
The expression that allows to check the count of log lines matching given expression could be used for checking if particular error did not ocurred during the test.

The syntax could be (as discussed with @pepoviola):
```
alice: count of log lines (containing|matching) "pattern" is <comparator> <number> [within]
```
e.g.:
```
alice: count of log lines containing "error" is  0 within 30s
```

`pest` grammer entry:
```
count_log_match = { node_name ~ "count of log lines" ~ ("containing"|"matching") ~ match_type? ~ double_quoted_string ~ "is" ~ (comparison | int+) ~ within? }
```

PR #603, opened 2022-11-17
Added assertion for matching lines count

michalkucharczyk:
This allows to add assertion for number of lines matchin given pattern, e.g:
```
alice: count of log lines containing "error" is 0 within 30s
```

The syntax is:
```
node_name: count of log lines (containing|matching) "pattern" is <comparator> <number> [within]
```

Currently the implementation waits `timeout` amount of time, and then captures and searches the log.
Assertions are not executed in parallel.

Fixes: #598

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -285,7 +285,7 @@ dependencies = [
 
 [[package]]
 name = "dsl-parser-wrapper"
-version = "0.1.6"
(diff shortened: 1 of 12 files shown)


pepoviola:
Looks great @michalkucharczyk!! thanks for your contribution 🙌

PR #625, opened 2022-12-04
allow to set use a raw chain-spec for paras under chain_spec_path key

### LINKED ISSUE #507: Feature: Support raw chain
It would be great if we can pass an already configured raw chain to the node (specially the parachain ones). 
Zombienet would not try to modify the genesis/state of the specs.

Use case: When replaying exported state (raw) from an existing chain

pepoviola:
- Allow use `raw` chain-spec files under the key `chain_spec_path` for paras.

Fix #507 .


wirednkod (approved):
lgtm

PR #627, opened 2022-12-04
fix release job

pepoviola:
Fix release job for `arm` (now linux needs suffix)

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -28,7 +28,7 @@ jobs:
       - run: npm run package
         working-directory: "./javascript"
       - name: Test if zombienet-linux runs
-        run: ./bins/zombienet-linux version
+        run: ./bins/zombienet-linux-linux-x64 version
         working-directory: "./javascript"
       - name: release
         uses: actions/create-release@v1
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -28,7 +28,7 @@ jobs:
       - run: npm run package
         working-directory: "./javascript"
       - name: Test if zombienet-linux runs
-        run: ./bins/zombienet-linux version
+        run: ./bins/zombienet-linux-linux-x64 version
```
wirednkod, line 31: Should there be 2 `linux`?

.github/workflows/release.yaml
```diff
@@ -47,7 +47,7 @@ jobs:
         with:
           upload_url: ${{ steps.create_release.outputs.upload_url }}
           asset_path: ./javascript/bins/zombienet-linux-x64
-          asset_name: zombienet-linux
+          asset_name: zombienet-linux-linux-x64
```
wirednkod, line 50: Same as above

pepoviola: No, my bad. Nice catch 😀

PR #633, opened 2022-12-06
refactor `wait_node_ready` for native provider

### LINKED ISSUE #516: Do not rely on the logs to check if node is running
Rely on the the rcp log line is error prune, for example when the log line change (as already happens) we need to update our code to match the new line. We can check if the process is up using prometheus or we can just check if the log lines grow.

pepoviola:
- refactor the logic to verify that the node is ready:
  - First check if the `pid`   is running
  - Check if the logs lines grow
  
  fix #516


wirednkod (approved):
A nip tuck, not really necessary - just makes the code more readable. Feel free to discard. Other than that - looks good

PR #655, opened 2022-12-13
Add yaml parser

### LINKED ISSUE #606: Add `yaml` suport in the `cli` module
Imho using yaml instead of toml would make much more sense:
1. easier to map json to yaml (because json is subset of yaml)
2. k8s and docker are yaml
3. it can express heterogenous arrays
4. people in ops are good yaml and its templating

what do you think?

wirednkod:
Fixes #606


pepoviola (approved):
Nice!! thanks!!

PR #666, opened 2022-12-16
always print logs

### LINKED ISSUE #665: Print loki URLs on successful run too
The links to loki were previously printed always if I remember correctly.
Recently they seem to be only available for failed runs.

This is a bit inconvenient, because it can be super useful to verify successful logs too:
- Check successful logs against failing ones
- Verify a successful test did what you think it did

To do this I currently need to manually query loki via the identifier printed.

pepoviola:
fix #665 
- Print always path/command for logs.

### DIFF
--- a/javascript/packages/orchestrator/src/test-runner/index.ts
+++ b/javascript/packages/orchestrator/src/test-runner/index.ts
@@ -145,7 +145,7 @@ export async function run(
               if (inCI) {
                 // keep pods running for 30 mins.
                 console.log(
-                  `\n\t${decorators.yellow(
+                  `\n\t${decorators.red(
                     "One or more test failed, we will keep the namespace up for 30 more minutes",
                   )}`,
                 );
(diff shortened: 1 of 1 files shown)


javascript/packages/orchestrator/src/test-runner/index.ts
```diff
@@ -125,83 +125,99 @@ export async function run(
...
               await network.stop();
               break;
             case "kubernetes":
               if (inCI) {
-                // show links to grafana and also we need to move the logs to artifacts
-                const networkEndtime = new Date().getTime();
-                for (const node of network.relay) {
+                // keep pods running for 30 mins.
+                console.log(
+                  `\n\t${decorators.yellow(
```
wirednkod, line 150: Lets do that .red since its a failure

wirednkod (approved):
Other than a nip tuck and the ci failing it looks good to me

Issue #640, opened 2022-12-09
Support large genesis file


crystalin:
If the genesis is too big, you can't use it with ZombieNet:
`RangeError [ERR_FS_FILE_TOO_LARGE]: File size (8689689103) is greater than 2 GiB`

(You can take https://s3.us-east-2.amazonaws.com/snapshots.moonbeam.network/moonriver/latest/moonriver-state.json for ex)

pepoviola:
Hi @crystalin, thanks for reporting. This is because we read the json into memory. We can implement `streams` or also check if the file is in raw mode and only make the needed replacements. I will work in the fix and ping you for testing.

Thanks!!

PR #679, opened 2023-01-09
Fix big raw

pepoviola:
Fix https://github.com/paritytech/zombienet/issues/640

- Allow to use large genesis files (more than 2 GB), we can't customize in this case *but* this will be mostly `raw` files.
-  Add JSONStream `dep` and `fns` to check if the file is `raw` and get the `id`.

### DIFF
--- a/javascript/packages/orchestrator/src/chain-spec.ts
+++ b/javascript/packages/orchestrator/src/chain-spec.ts
@@ -543,7 +543,7 @@ export async function isRawSpec(specPath: string): Promise<boolean> {
 }
 
 export async function getChainIdFromSpec(specPath: string): Promise<string> {
-  return new Promise((res, _rej) => {
+  return new Promise((res) => {
     const stream = fs.createReadStream(specPath, { encoding: "utf8" });
     const parser = JSONStream.parse(["id"]);
     stream.pipe(parser);
--- a/javascript/packages/orchestrator/src/chain-spec.ts
+++ b/javascript/packages/orchestrator/src/chain-spec.ts
@@ -13,6 +13,8 @@ import { ChainSpec, HrmpChannelsConfig, Node } from "./types";
 const JSONbig = require("json-bigint")({ useNativeBigInt: true });
 const debug = require("debug")("zombie::chain-spec");
 
+const JSONStream = require("JSONStream");
+
 // track 1st staking as default;
 let stakingBond: number | undefined;
 
@@ -374,7 +376,21 @@ export async function changeGenesisConfig(specPath: string, updates: any) {
 }
 
 export async function addBootNodes(specPath: string, addresses: string[]) {
-  const chainSpec = readAndParseChainSpec(specPath);
+  let chainSpec;
+  try {
+    chainSpec = readAndParseChainSpec(specPath);
+  } catch(e: any) {
+    if(e.code !== "ERR_FS_FILE_TOO_LARGE") throw e;
+
+    // can't customize bootnodes
+    console.log(
+      `\n\t\t 🚧 ${decorators.yellow(
+        `Chain Spec file ${specPath} is TOO LARGE to customize (more than 2G).`,
+      )} 🚧`,
+    );
+    return;
+  }
+
   // prevent dups bootnodes
   chainSpec.bootNodes = [...new Set(addresses)];
   writeChainSpec(specPath, chainSpec);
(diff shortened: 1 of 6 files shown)


javascript/packages/orchestrator/src/chain-spec.ts
```diff
@@ -13,6 +13,8 @@ import { ChainSpec, HrmpChannelsConfig, Node } from "./types";
 const JSONbig = require("json-bigint")({ useNativeBigInt: true });
 const debug = require("debug")("zombie::chain-spec");
 
+const JSONStream = require("JSONStream");
```
wirednkod, line 16: For consistency reasons lets use import here:

```suggestion
import JSONStream  from "JSONStream";
```

javascript/packages/orchestrator/src/chain-spec.ts
```diff
@@ -374,7 +376,21 @@ export async function changeGenesisConfig(specPath: string, updates: any) {
...
 export async function addBootNodes(specPath: string, addresses: string[]) {
-  const chainSpec = readAndParseChainSpec(specPath);
+  let chainSpec;
+  try {
+    chainSpec = readAndParseChainSpec(specPath);
+  } catch (e: any) {
+    if (e.code !== "ERR_FS_FILE_TOO_LARGE") throw e;
+
+    // can't customize bootnodes
+    console.log(
```
wirednkod, line 386: Since this is an error, it should be red instead of yellow, and maybe we should also print it

javascript/packages/orchestrator/src/chain-spec.ts
```diff
@@ -510,6 +526,38 @@ export function writeChainSpec(specPath: string, chainSpec: any) {
   }
 }
 
+export async function isRawSpec(specPath: string): Promise<boolean> {
+  return new Promise((res, _rej) => {
```
wirednkod, line 530:
```suggestion
  return new Promise((res) => {
```

javascript/packages/orchestrator/src/chain-spec.ts
```diff
@@ -510,6 +526,38 @@ export function writeChainSpec(specPath: string, chainSpec: any) {
...
+      return res(true);
+    });
+    stream.on("end", () => {
+      return res(false);
+    });
+  });
+}
+
+export async function getChainIdFromSpec(specPath: string): Promise<string> {
+  return new Promise((res, _rej) => {
```
wirednkod, line 546:
```suggestion
  return new Promise((res) => {
```

javascript/packages/orchestrator/src/paras.ts
```diff
@@ -90,20 +90,20 @@ export async function generateParachainFiles(
...
-    const relayChainSpec = readAndParseChainSpec(relayChainSpecFullPathPlain);
-    if (plainData.para_id) plainData.para_id = parachain.id;
-    if (plainData.paraId) plainData.paraId = parachain.id;
-    if (plainData.relay_chain) plainData.relay_chain = relayChainSpec.id;
-    if (plainData.genesis.runtime?.parachainInfo?.parachainId)
-      plainData.genesis.runtime.parachainInfo.parachainId = parachain.id;
-
-    writeChainSpec(chainSpecFullPathPlain, plainData);
-
+    //const plainData = readAndParseChainSpec(chainSpecFullPathPlain);
```
wirednkod, line 93: if this comment is not needed code lets remove it

javascript/packages/orchestrator/src/paras.ts
```diff
@@ -163,11 +163,22 @@ export async function generateParachainFiles(
...
+      // ensure the correct para_id
+      const paraSpecRaw = readAndParseChainSpec(chainSpecFullPath);
+      if (paraSpecRaw.para_id) paraSpecRaw.para_id = parachain.id;
+      if (paraSpecRaw.paraId) paraSpecRaw.paraId = parachain.id;
+      writeChainSpec(chainSpecFullPath, paraSpecRaw);
+    } catch (e: any) {
+      if (e.code !== "ERR_FS_FILE_TOO_LARGE") throw e;
+
+      // can't customize para_id
+      console.log(
```
wirednkod, line 176: same comment as above

wirednkod (approved):
Some small nip tucks. Other than that it lgtm

javascript/packages/orchestrator/src/chain-spec.ts
pepoviola: Is not an error, the file is too big so we can't customize but we can deploy the namespace.

javascript/packages/orchestrator/src/paras.ts
pepoviola: Same here, just a warning.

Issue #664, opened 2022-12-14
What this error means `⚠ Can not access native, please check your config`  ?


dzmitry-lahoda:
I am trying to run native nix in docker-compose which builds images inline (as docker launches).

I see this error
```
,"types":{"Header":{"number":"u64","parent_hash":"Hash","post_state":"Hash"}}}
composable-devnet-1-1                      | > zombienet@0.1.0 zombie
composable-devnet-1-1                      | > node ./packages/cli/dist/cli.js spawn /devnet-dali-centauri-1.json
composable-devnet-1-1                      | 
composable-hyperspace-create-clients-1 exited with code 0
composable-devnet-2-1 exited with code 1
composable-devnet-1-1                      | ╔════════════════════╤════════════════════════════════════════════════════════════════════════════════════════════════════╗
composable-devnet-1-1                      | ║ 🧟 Zombienet 🧟    │ Initiation                                                                                         ║
composable-devnet-1-1                      | ╟────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────╢
composable-devnet-1-1                      | ║ Provider           │ native                                                                                             ║
composable-devnet-1-1                      | ╟────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────╢
composable-devnet-1-1                      | ║ Namespace          │ zombie-84bac59ff5fa3573155f9272a0c53190                                                            ║
composable-devnet-1-1                      | ╟────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────╢
composable-devnet-1-1                      | ║ Temp Dir           │ /tmp/zombie-84bac59ff5fa3573155f9272a0c53190_-24-x4Q15glzuouw                                      ║
composable-devnet-1-1                      | ╚════════════════════╧════════════════════════════════════════════════════════════════════════════════════════════════════╝
composable-devnet-1-1                      | 
composable-devnet-1-1                      |             ⚠ Can not access native, please check your config.
composable-devnet-1-1                      | npm notice 
composable-devnet-1-1                      | npm notice New major version of npm available! 8.19.2 -> 9.2.0
composable-devnet-1-1                      | npm notice Changelog: <https://github.com/npm/cli/releases/tag/v9.2.0>
composable-devnet-1-1                      | npm notice Run `npm install -g npm@9.2.0` to update!
composable-devnet-1-1                      | npm notice 
```

What this errors mean?

[...]

PR #683, opened 2023-01-10
Which bins fix

dzmitry-lahoda:
Use default bins for native. Which does no work, neither with "" around not raw. Nixos stuff.

https://github.com/paritytech/zombienet/issues/664


pepoviola (approved):
Looks good, thanks!!

wirednkod (approved):

pepoviola:
thanks @dzmitry-lahoda!!

PR #702, opened 2023-01-27
add logic for allow `--listen-addr` in args

pepoviola:
- This only allow to use it in relay-chain nodes.

### DIFF
--- a/javascript/packages/orchestrator/src/cmdGenerator.ts
+++ b/javascript/packages/orchestrator/src/cmdGenerator.ts
@@ -286,15 +286,14 @@ export async function genCmd(
 
   const listenIndex = args.findIndex((arg) => arg === "--listen-addr");
   if (listenIndex >= 0) {
-    let listenAddrParts = args[listenIndex+1].split("/");
+    let listenAddrParts = args[listenIndex + 1].split("/");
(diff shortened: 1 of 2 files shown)


javascript/packages/orchestrator/src/cmdGenerator.ts
```diff
@@ -283,7 +283,17 @@ export async function genCmd(
   for (const [k, v] of Object.entries(portFlags)) {
     args.push(...[k, v.toString()]);
   }
-  args.push(...["--listen-addr", `/ip4/0.0.0.0/tcp/${nodeSetup.p2pPort}/ws`]);
+
+  const listenIndex = args.findIndex((arg) => arg === "--listen-addr");
+  if (listenIndex >= 0) {
+    let listenAddrParts = args[listenIndex + 1].split("/");
+    listenAddrParts[listenAddrParts.length - 2] = `${nodeSetup.p2pPort}`;
```
melekes, line 290:
```
listenAddrParts[3] = `${nodeSetup.p2pPort}`;
```

would've been less error prone because addr can contain `/p2p/SDSvsdfsdfsdfsdfsdf` in the end.

pepoviola: Yes, make sense. I will change it asap.
Thanks!!

PR #708, opened 2023-01-31
Auth-service golang app PR

emamihe:
Hi @pepoviola 
This is the auth service written in go
This will help us that zombienet-cli can use restricted tokens into specific zombie-xyz namespace to do the setup of simulators to have much more secure access to the cluster


pepoviola:
Hi @emamihe, looks good in general. Can we coordinate a walkthrough to check how we can integrate with the cli.

Thanks!

pepoviola:
@emamihe is this the updated version of the auth service?

PR #731, opened 2023-02-10
Fix paraNodes not being added to the logs dir

### LINKED ISSUE #699: Artifact logs do not contain collator logs
When a job is ran, zombienet states `📓 Logs are also available in the artifacts' pipeline in gitlab`.
When I inspect the artifacts, I see only the ones from the validator however, no collator logs are included.

[Example test](https://gitlab.parity.io/parity/mirrors/cumulus/-/jobs/2307678)

All nodes as listed
image

Artifacts
image

skunert:
Closes #699

Small fix, logs were not dumped in `logs` directory because `concat` returns a new array instead of modifying `memo`.
Now artifact logs for collators should also show up in CI.

### DIFF
--- a/javascript/package-lock.json
+++ b/javascript/package-lock.json
@@ -4497,11 +4497,11 @@
     },
     "packages/cli": {
       "name": "@zombienet/cli",
-      "version": "1.3.30",
+      "version": "1.3.34",
       "license": "GPL-3.0-or-later",
       "dependencies": {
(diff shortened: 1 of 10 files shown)


pepoviola (approved):
Nice! Thanks!!

Can you run:

npm run lint:write 

To lint the code, thanks!

pepoviola:
thanks @skunert !! 🙌

PR #741, opened 2023-02-20
Fix minor rust warnings

wirednkod:
Fix [minor warnings](https://github.com/paritytech/zombienet/pull/708/files#file-crates-parser-src-lib-rs-L40) that are appearing when a PR is opened.


pepoviola (approved):

pepoviola:
Looks good! thanks!

PR #750, opened 2023-02-22
bump for release 1.3.35

pepoviola:
Bump for new release.

### DIFF
--- a/javascript/packages/cli/package.json
+++ b/javascript/packages/cli/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@zombienet/cli",
-  "version": "1.3.34",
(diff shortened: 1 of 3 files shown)


wirednkod:
You need first to release the utils 0.15 I think

pepoviola:
> You need first to release the utils 0.15 I think

but utils@0.15 is bumped also.

wirednkod:
> > You need first to release the utils 0.15 I think
> 
> but utils@0.15 is bumped also.

During build the 0.15 of utils is not found in the npm and thus fails. I would propose, first to create a PR that will release 0.15 and then make the release

wirednkod (approved):

PR #755, opened 2023-02-24
fix for undying/adder collator

pepoviola:
- undying/adder collators command were generated without the `--prometheus-extenal` flag, so checking if those `are up` wasn't possible.
- 
Thanks!

### DIFF
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -695,10 +695,9 @@ async function getExternalPorts(
   return ports;
 }
 
-
-  // enable --prometheus-external by default
-  const prometheusExternal = (networkSpec: any): boolean => {
-    return networkSpec.settings?.prometheus !== undefined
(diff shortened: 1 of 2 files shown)


javascript/packages/orchestrator/src/configGenerator.ts
```diff
@@ -699,3 +694,10 @@ async function getExternalPorts(
 
   return ports;
 }
+
+// enable --prometheus-external by default
+const prometheusExternal = (networkSpec: any): boolean => {
```
wirednkod, line 699:
```suggestion
// TODO: fix the `any` to an actual interface
const prometheusExternal = (networkSpec: any): boolean => {
```

wirednkod (approved):
just a todo comment. Other than that looks good

javascript/packages/orchestrator/src/configGenerator.ts
pepoviola: Yes, better to track this as `TODO`.
Thanks!

PR #762, opened 2023-02-28
Adjust first set of examples

wirednkod:
Adjust the examples in order to remove the adder-collator and add polkadot-parachain

PR #768, opened 2023-03-03
Replace adder-collator with polkadot-parachain as default

wirednkod:


wirednkod:
Close here - will make this fix in part of the #762 PR

PR #770, opened 2023-03-03
Error when using -d option

### LINKED ISSUE #769: Error when using `-d` option
When `-d` option is used with CLI the following error occurs:
```bash
 Error:  	 Error: Cannot find module './firstnet/polkadot-local-plain.json'
Require stack:
- /snapshot/javascript/packages/orchestrator/dist/orchestrator.js
- /snapshot/javascript/packages/orchestrator/dist/index.js
- /snapshot/javascript/packages/cli/dist/cli.js
1) If you want to compile the package/file into executable, please pay attention to compilation warnings and specify a literal in 'require' call. 2) If you don't want to compile the package/file into executable and want to 'require' it from filesystem (likely plugin), specify an absolute path in 'require' call using process.cwd() or process.execPath.
```

The `toml` file that used is of a simple network:
```toml
[settings]
provider = "native"
timeout = 1000

[relaychain]
chain = "polkadot-local"

  [[relaychain.nodes]]
  name = "alice"

  [[relaychain.nodes]]
  name = "bob"
```

image

wirednkod:
Fixes #769


javascript/packages/orchestrator/src/orchestrator.ts
```diff
@@ -261,7 +261,8 @@ export async function start(
 
     // Check if the chain spec is in raw format
     // Could be if the chain_spec_path was set
-    const chainSpecContent = require(chainSpecFullPathPlain);
+    const chainSpecContentTest = fs.readFileSync(chainSpecFullPathPlain);
+    const chainSpecContent = JSON.parse(chainSpecContentTest.toString());
```
pepoviola, line 265:
```suggestion
     const chainSpecContent = readAndParseChainSpec(chainSpecFullPathPlain);
```

javascript/packages/orchestrator/src/orchestrator.ts
```diff
@@ -375,7 +376,8 @@ export async function start(
 
     // ensure chain raw is ok
     try {
-      const chainRawContent = require(chainSpecFullPath);
+      const chainsSpecString = fs.readFileSync(chainSpecFullPath).toString();
+      const chainRawContent = JSON.parse(chainsSpecString);
```
pepoviola, line 380:
```suggestion
```suggestion
     const chainSpecContent = readAndParseChainSpec(chainSpecFullPathPlain);
```
```

File: javascript/packages/orchestrator/src/providers/k8s/chain-spec.ts [hunk 3]
```diff
@@ -8,7 +8,7 @@ import { getClient } from "../client";
 import { createTempNodeDef, genNodeDef } from "./dynResourceDefinition";
 const debug = require("debug")("zombie::kube::chain-spec");
 
-const fs = require("fs").promises;
+const { writeFileSync, promises } = require("fs");
```
pepoviola, line 11: Can we rename `promises`? e.g somethig less generic.
```suggestion
const { writeFileSync, promises as fsPromises } = require("fs");
```

javascript/packages/orchestrator/src/providers/native/chain-spec.ts
```diff
@@ -90,7 +90,7 @@ export async function getChainSpecRaw(
   // let's add some extra checks here to ensure we are ok.
   let isValid = false;
   try {
-    require(chainFullPath);
+    JSON.parse(readFileSync(chainFullPath).toString());
```
pepoviola, line 93:
```suggestion
    readAndParseChainSpec(chainFullPath);
```

wirednkod:
There is a minor issue concerning the podman - that still makes the `-d` option fail. It is relative to the permissions of the user which are not the same between the local to the podman directories.
I am not also sure, why we need to create everything locally and then "pass" it to the podman, and not create it inside podman - opinions @pepoviola ?

pepoviola:
> There is a minor issue concerning the podman - that still makes the `-d` option fail. It is relative to the permissions of the user which are not the same between the local to the podman directories. I am not also sure, why we need to create everything locally and then "pass" it to the podman, and not create it inside podman - opinions @pepoviola ?

Fixed in the last commit, using `podman unshare` to change permissions.

pepoviola (approved):

PR #790, opened 2023-03-14
Feat add silent cli

### LINKED ISSUE #787: Add option to disable logs in `@zombienet/orchestrator`
I'm attempting to use the programmatic API of `@zombienet/orchestrator` in Capi, and it works almost perfectly... except that it makes a lot of logs that we want to suppress. As a result, instead of being able to simply write
```ts
const network = await start("", config, options)
return network
```
We would have to instead write
```ts
const child = new Deno.Command(Deno.execPath(), {
  args: [
    "run",
    "-A",
    import.meta.resolve("./zombienet_worker.ts"),
    JSON.stringify([config, options]),
  ],
  stdout: "piped",
  stderr: "inherit",
}).spawn()
for await (const line of readLines(readerFromStreamReader(child.stdout.getReader()))) {
  if (!line.startsWith("capi_network = ")) continue
  const network = JSON.parse(line.slice("capi_network = ".length)) as Network
  return network
}

// zombienet_worker.ts
import { LaunchConfig, start } from "npm:@zombienet/orchestrator"

const [config, options] = JSON.parse(Deno.args[0]!) as [LaunchConfig, Parameters<typeof start>[2]]

const network = await start("", config, options)

console.log(`capi_network = ${JSON.stringify(network)}`)
```
This is far from ideal, but it would greatly simplify our zombienet integration in capi if we could bypass the CLI and use the API directly.

I see no way to disable the logs from `@zombienet/orchestrator` without modifying `console.log`, which would affect everything else in the realm. Ideally, the logs could be disabled by passing a parameter to the `start` function (or similar).

pepoviola:
fix #787 

Allow to set `silent` (true by default in orchestrator) and enabled by default in the `cli`.
So, when you use the cli you can use the `-s` flag to suppress the output to stdout and when you use as library by default we don't log anything to stdout.

Thanks!

### DIFF
--- a/javascript/packages/cli/src/cli.ts
+++ b/javascript/packages/cli/src/cli.ts
@@ -418,7 +418,6 @@ async function spawn(
   // since this shouldn't be a bottleneck in most of the cases,
   // but also can be set with the `-c` flag.
   const spawnConcurrency = opts.spawnConcurrency || 4;
-
   const configPath = resolve(process.cwd(), configFile);
   if (!fs.existsSync(configPath)) {
     console.error(


javascript/packages/cli/src/cli.ts
```diff
@@ -344,7 +345,8 @@ program
       "Directory path for placing the network files instead of random temp one (e.g. -d /home/user/my-zombienet)",
     ),
   )
-  .addOption(new Option("-f, --force", "Force override all prompt commands"));
+  .addOption(new Option("-f, --force", "Force override all prompt commands"))
+  .addOption(new Option("-s, --silent", "Don't log to stdout"));
```
wirednkod, line 349:
```suggestion
  .addOption(new Option("-s, --silent", "Mute logging output."));
```

javascript/packages/cli/src/cli.ts
```diff
@@ -418,6 +420,8 @@ async function spawn(
   // since this shouldn't be a bottleneck in most of the cases,
   // but also can be set with the `-c` flag.
   const spawnConcurrency = opts.spawnConcurrency || 4;
+  const silent = opts.silent || false;
```
wirednkod, line 423:
```suggestion
  const silent = !!opts?.silent;
```

javascript/packages/cli/src/cli.ts
```diff
@@ -519,13 +523,17 @@ async function test(
     process.exit(1);
   }
 
+  const silent = opts.silent || false;
```
wirednkod, line 526:
```suggestion
  const silent = !!opts?.silent;
```

wirednkod (changes requested):
I find the idea of adding in a CLI the option to be silent is not great. The code placed here is not having a great overhead to the existing code.
Since we are "serving" the CLI also as a library, I suggest to alter the output/return of the code in to 3 different options:
a) when default (no flags) are set, then the return should be as is;
b) when "silent" set, then no output will be returned at all ("library" or not);
c) when "library" set, return information (not logs/table) of the "current status"
 
 I think we can merge this, but also a new `issue` should open, that will describe the above for implementation;

javascript/packages/cli/src/cli.ts
pepoviola: `opts` will always exists here since is created by `program.opts()` above.

pepoviola:
> I find the idea of adding in a CLI the option to be silent is not great. The code placed here is not having a great overhead to the existing code. Since we are "serving" the CLI also as a library, I suggest to alter the output/return of the code in to 3 different options: a) when default (no flags) are set, then the return should be as is; b) when "silent" set, then no output will be returned at all ("library" or not); c) when "library" set, return information (not logs/table) of the "current status"
> 
> I think we can merge this, but also a new `issue` should open, that will describe the above for implementation;

Wait, we are not serving the `cli` as a library. The `cli` module doesn't export any function that can be used from a caller. Is just an `entrypoint` to run zombienet and interact with the orchestrator.

wirednkod:
> Wait, we are not serving the `cli` as a library. The `cli` module doesn't export any function that can be used from a caller. Is just an `entrypoint` to run zombienet and interact with the orchestrator.

Yes - sorry for not clarifying - I meant actually the `orchestrator` and not the `cli` here. What I meant to say is that it makes no sense to run the CLI without any output. It is better to add this as a param variable of the orchestrator that will be always false for the CLI

javascript/packages/cli/src/cli.ts
```diff
@@ -32,6 +33,8 @@ import {
   DEFAULT_PROVIDER,
 } from "./constants";
 
+// We always want to log when the orchestrator is called from the cli
+const SILENT = false;
```
wirednkod, line 37: No reason for the CLI to even know about the "silent" or for this to exist 
```suggestion
```

javascript/packages/cli/src/cli.ts
```diff
@@ -418,6 +421,8 @@ async function spawn(
   // since this shouldn't be a bottleneck in most of the cases,
   // but also can be set with the `-c` flag.
   const spawnConcurrency = opts.spawnConcurrency || 4;
+
+  setSilent(SILENT);
```
wirednkod, line 425: Same as above:
```suggestion
```

javascript/packages/cli/src/cli.ts
```diff
@@ -519,13 +531,16 @@ async function test(
     process.exit(1);
   }
 
+  setSilent(SILENT);
+
```
wirednkod, line 535:
```suggestion
```

javascript/packages/cli/src/cli.ts
```diff
@@ -519,13 +531,16 @@ async function test(
...
   await run(
     configBasePath,
     testName,
     testDef,
     providerToUse,
     inCI,
     opts.spawnConcurrency,
+    SILENT,
```
wirednkod, line 543:
```suggestion
    false,
```

wirednkod (changes requested):

javascript/packages/cli/src/cli.ts
pepoviola: It seems reasonable, let me refactor this. Thanks!

javascript/packages/cli/src/cli.ts
```diff
@@ -418,6 +418,7 @@ async function spawn(
   // since this shouldn't be a bottleneck in most of the cases,
   // but also can be set with the `-c` flag.
   const spawnConcurrency = opts.spawnConcurrency || 4;
+
```
pepoviola, line 421:
```suggestion

```

wirednkod (approved):

Issue #788, opened 2023-03-14
Starting multiple networks in tests


serban300:
We would like to write zombienet tests for the  [bridges project](https://github.com/paritytech/parity-bridges-common). Since a bridge connects 2 networks we would need the ability to start multiple networks as part of `zndsl` test.

If I understand correctly, right now a `zndsl` test definition supports only one network. I was wondering if it would be possible to add support for multiple networks.

If the proposal makes sense and if it's ok for you, I would be happy to implement this feature myself. I was thinking of the following very high-level approach:

1. Allowing multiple `Network: network_def.json` lines in the `.zndsl` file and building a vector of networks when parsing it instead of a single network. 
2. I think we should assume that each network uses different node names. This would help preserve the current syntax and keep the code changes minimal.
3. In the orchestrator iterating through the vector of networks and starting each one.
4. I see that for the moment there is only one `Client` per `zndsl` run. We would probably need one per network. I'm not completely sure about this, but I guess there must be a solution.

pepoviola:
Hi @serban300, thanks for the feedback, open this issue and the draft PR :) 🙌 .

We will open this new feature as discussion with the team and get back to you asap. I like the idea of running multiple-networks what I wonder what is the best api/config for users take advantage of this.

Thanks!!

cc @wirednkod

PR #792, opened 2023-03-15
`zndsl`: support multiple networks

serban300:
Just a draft implementation to showcase what we would need for #788

Might be missing some corner cases and might need some polishing

### DIFF
--- a/crates/parser/src/ast.rs
+++ b/crates/parser/src/ast.rs
@@ -147,7 +147,7 @@ pub struct Assertion {
 #[derive(Debug, Serialize, Deserialize, PartialEq, Eq)]
 pub struct TestDefinition {
     pub description: Option<String>,
-    pub network: String,
+    pub networks: Vec<String>,
     pub creds: String,
     pub assertions: Vec<Assertion>,
 }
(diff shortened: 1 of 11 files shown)


wirednkod:
Thank you @serban300 for this. I have "cloned" it and added some more lines (see #796 ) to fix the breaking CIs and lints. Still though the tests are not working.
I think it is a great idea in order to make the CLI work, to use instances of the orchestrator in order to run each network. You can either continue woring on my PR or copy/paste the changes I made in order to make yours work, and we can continue from there.

serban300:
@wirednkod Thank you for the fixes and the suggestion. I will try to implement it and fix the remaining CI failures.

wirednkod:
> @wirednkod Thank you for the fixes and the suggestion. I will try to implement it and fix the remaining CI failures.

As mentioned in my previous comment: I think it is better to `use instances of the orchestrator in order to run each network`. 

Just FYI I will try to put together this kind of approach in a different PR - in order to wrap my mind around your suggestion. Again thank you for the suggestion - it rings as a great idea.

serban300:
@wirednkod I didn't get to experiment with different instances of the orchestrator yet. I was trying to understand the CI failures first, and I think they occur because the `zombie` executable probably uses a version of `parser-wrapper` which is published from the latest `main` branch and not from this PR. Do you have a process for handling this kind of situations, where there is a change in the `parser` ? I didn't find anything in the documentation. This PR would need to use a locally compiled version of `@zombienet/dsl-parser-wrapper`.

serban300:
Closing in favor of #796

PR #796, opened 2023-03-15
`zndsl:` support multiple networks

wirednkod:
Added commit ontop of https://github.com/paritytech/zombienet/pull/792 for checking if build works


serban300:
Temporarily added a commit for using a locally compiled version of the `dsl-parser-wrapper` just to check if it fixes the CLI tests. Will revert it later.

serban300:
I think I covered all the corner cases. The PR should be ready for review.

Sorry, I force pushed a couple of times to make it easier to follow.

Notes:
1. I added a [commit](https://github.com/paritytech/zombienet/pull/796/commits/d702e06ed84d586af7d96fb1db64499fad33465a) for using a locally compiled version of `dsl-parser-wrapper`. That should be removed, but not sure what would be the procedure to publish a new version of `dsl-parser-wrapper`. So leaving it like this for the moment.
2. I'm not sure if I understood what using multiple instances of the orchestrator would suppose. @wirednkod could you add more details on that idea please ? 

@pepoviola @wirednkod could you PTAL when you have time ?

wirednkod:
@serban300 there are some conflicts on this - maybe you can have a look?

serban300:
> @serban300 there are some conflicts on this - maybe you can have a look?

Sure, I'll rebase it these days as soon as I have some spare time

serban300:
> @serban300 there are some conflicts on this - maybe you can have a look?

@wirednkod Done

pepoviola:
Hi @serban300, thanks for this pr. We talked internally and we agree and this kind of meta-orchestration is better to be handled externally. I think will be better to have a new `cli` binary that allow users to spawn/test multiple instance of zombienet networks, so this new cli tool can oversee the `n` networks. I can move this pr to this new `cli` and work on the `dx` to make it easy to use.

 I will ping you to make you a couple of question related to your use case in order to have a clear picture of your use case.
Thanks for your feedback and to bring us a new use case and the code to support it 🙌

serban300:
@pepoviola thank you for considering this use-case and please feel free to ping me.

Having a new cli tool sounds good. Please keep us up to date on it. Is there an issue or a repo that we could follow ?

pepoviola:
> @pepoviola thank you for considering this use-case and please feel free to ping me.
> 
> Having a new cli tool sounds good. Please keep us up to date on it. Is there an issue or a repo that we could follow ?

Hi @serban300, sorry about the delay. Thanks again for your feedback. Is your use-case require the test-runner through the DSL? I mean, what I think could be good is a to have a `puppeteer` process that internally spawn `n` zombienet processes and keep track to the `state` and endpoints, so you can then interact with each network. **But** if you want to use the `DSL` with assertions for `nodes` in different networks we should find a way to get an unique reference to each one.

I will ping you in element to check if this is useful to you.
Thanks!

svyatonik:
> @svyatonik @bkontur what do you think ? Do you have some specific test cases in mind or any input here ?

Apart from simple CI tests, we'll probably want to use similar tests for checking how bridge behaves on runtime upgrades. I assume there's a way to test runtime upgrades in zombienet, but additionally it'd be nice to have an option e.g. to stop external processes that @serban300 has mentioned when certain events (for us: runtime upgrade) happens and start some other process. Something like:
- start networks
- start external process (bridge relay)
- submit some transactions (do we also need a separate process for that?)
- wait for some blocks
- do some external assertions (some messages are delivered)
- stop our external process
- upgrade selected chain
- start other external process
- wait for some blocks
- do some external assertions (some messages are delivered)

bkontur:
> @svyatonik @bkontur what do you think ? Do you have some specific test cases in mind or any input here ?

for CI - zombienet with more networks and ability to send/check transactions everywhere, then we can do whatever inegration test.
And exactly like Slava wrote about runtime upgrade, there is a older issue for testing runtime upgrade in Cumulus: https://github.com/paritytech/cumulus/issues/2264

PR #845, opened 2023-03-24
Replace axios with fetch API in `cli` package

### LINKED ISSUE #846: Replace Axios with Fetch in CLI package;

wirednkod:
- Replace `axios` with `fetch` API in CLI;

- Replace npm package `progress` with `cli-progress`:

image

Fixes #846

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -15,7 +15,7 @@ stages:
   - deploy
 
 variables:
-  CI_IMAGE: "node:16.18"
+  CI_IMAGE: "node:18"
   BUILDAH_IMAGE: "quay.io/buildah/stable:v1.27"
   DOCKERFILE: scripts/ci/docker/zombienet_injected.Dockerfile
   DOCKERHUB_REPO: paritytech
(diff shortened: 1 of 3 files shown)


javascript/packages/cli/src/cli.ts
```diff
@@ -49,52 +48,81 @@ const downloadBinaries = async (binaries: string[]): Promise<void> => {
...
+                  });
+                  writer.write(res.value);
+                  read();
+                }
+              }
+            } catch (error) {
+              console.error(error);
+              reject(error);
+            }
+          }
```
l0r1s, line 114: The naming of the function is not clear and it's using many variabe from outer scope, making it hard to extract it and bloating the outer function a bit more, why use recursion in that case ?

pepoviola: I think will be more clear to use the same patter that you use in the `utils` package @wirednkod with the `while` loop and `break`ing when is done.

wirednkod: Thank you both - Will address it asap

pepoviola (approved):

PR #851, opened 2023-03-24
Minor Fix on the docker image for example 0003

wirednkod:


examples/0003-big-network.toml
```diff
@@ -1,5 +1,5 @@
 [relaychain]
-default_image = "docker.io/paritypr/polkadot:latest"
+default_image = "docker.io/parity/polkadot:latest"
```
l0r1s, line 2: Good catch 🎯

l0r1s (approved):

PR #867, opened 2023-03-27
[Draft]: Fix podman v4 and refactor

l0r1s:
Preparation work to support podman v4 and refactoring the podman provider to make it cleaner.

The different resources have been split on individual file exposing only one public method as an API to generate the specification from the parameter given in the constructor, hence, respecting the [Single Responsibility Principle](https://en.wikipedia.org/wiki/Single-responsibility_principle).

The issues with the v4 will be the naming of pod which are not suffixed with _pod anymore if the pod has the same name as the container inside it.

The short term objective is to make the code cleaner and the medium term one is to be able to inverse dependency on filesystem/api call/shell ([Dependency inversion principle](https://en.wikipedia.org/wiki/Dependency_inversion_principle)) call to make it easier to unit test providers in an isolated fashion, starting with the podman provider.

Open to suggestions.

### DIFF
--- a/javascript/packages/orchestrator/src/providers/podman/resources/grafana.resource.ts
+++ b/javascript/packages/orchestrator/src/providers/podman/resources/grafana.resource.ts
@@ -1,0 +1,113 @@
+import { getRandomPort, makeDir } from "@zombienet/utils";
+import fs from "fs/promises";
+import path from "path";
+import { Client } from "../../client";
+
+export class GrafanaResource {
+  private readonly dataSourcesPath: string;
+
+  constructor(
+    client: Client,
+    private readonly namespace: string,
+    private readonly prometheusIp: string,
+    private readonly tempoIp: string,
+  ) {
+    this.dataSourcesPath = `${client.tmpDir}/grafana/datasources`;
+  }
+
+  public async generateSpec() {
+    const volumes = await this.generateVolumes();
+    const volumeMounts = this.generateVolumesMounts();
+    const containersPorts = await this.generateContainersPorts();
+    const containers = this.generateContainers(volumeMounts, containersPorts);
+    const podSpec = this.generatePodSpec(containers, volumes);
+
+    return podSpec;
+  }
+
+  private async createVolumeDirectories() {
+    await makeDir(this.dataSourcesPath, true);
+  }
+
+  private async generateGrafanaConfig() {
+    const templateConfigPath = path.resolve(__dirname, "./configs/grafana.yml");
+    const grafanaConfigBuffer = await fs.readFile(templateConfigPath);
+
+    let grafanaConfig = grafanaConfigBuffer.toString("utf8");
+    grafanaConfig = grafanaConfig
+      .replace("{{PROMETHEUS_IP}}", this.prometheusIp)
+      .replace("{{TEMPO_IP}}", this.tempoIp);
+
+    await fs.writeFile(`${this.dataSourcesPath}/prometheus.yml`, grafanaConfig);
+  }
+
+  private async generateVolumes() {
+    await this.createVolumeDirectories();
+    await this.generateGrafanaConfig();
+
+    const dataSourcesVolume = {
+      name: "datasources-cfg",
+      hostPath: { type: "Directory", path: this.dataSourcesPath },
+    };
+
+    return [dataSourcesVolume];
+  }
+
+  private generateVolumesMounts() {
+    const dataSourcesVolumeMount = {
+      name: "datasources-cfg",
+      mountPath: "/etc/grafana/provisioning/datasources",
+      readOnly: false,
+    };
+
+    return [dataSourcesVolumeMount];
+  }
+
+  private async generateContainersPorts() {
+    const grafanaWebPort = {
+      containerPort: 3000,
+      name: "grafana_web",
+      hostPort: await getRandomPort(),
+    };
+
+    return [grafanaWebPort];
+  }
+
+  private generateContainers(volumeMounts: any, ports: any) {
+    const grafanaContainerSpec = {
+      image: "docker.io/grafana/grafana",
+      name: "grafana",
+      imagePullPolicy: "Always",
+      ports,
+      volumeMounts,
+    };
+
+    return [grafanaContainerSpec];
+  }
+
+  private generatePodSpec(containers: any[], volumes: any[]) {
+    const grafanaPodSpec = {
+      apiVersion: "v1",
+      kind: "Pod",
+      metadata: {
+        name: "grafana_pod",
+        namespace: this.namespace,
+        labels: {
+          "app.kubernetes.io/name": this.namespace,
+          "app.kubernetes.io/instance": "grafana",
+          "zombie-role": "grafana",
+          app: "zombienet",
+          "zombie-ns": this.namespace,
+        },
+      },
+      spec: {
+        hostname: "grafana",
+        restartPolicy: "OnFailure",
+        volumes,
+        containers,
+      },
+    };
+
+    return grafanaPodSpec;
+  }
+}
--- a/javascript/packages/orchestrator/src/providers/podman/resources/introspector.resource.ts
+++ b/javascript/packages/orchestrator/src/providers/podman/resources/introspector.resource.ts
@@ -1,0 +1,65 @@
+import { getRandomPort } from "@zombienet/utils";
+import { INTROSPECTOR_POD_NAME } from "../../../constants";
+
+export class IntrospectorResource {
+  constructor(
+    private readonly namespace: string,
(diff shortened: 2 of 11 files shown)


javascript/packages/orchestrator/src/providers/podman/resources/grafana.resource.ts
```diff
@@ -0,0 +1,113 @@
...
+  }
+
+  public async generateSpec() {
+    const volumes = await this.generateVolumes();
+    const volumeMounts = this.generateVolumesMounts();
+    const containersPorts = await this.generateContainersPorts();
+    const containers = this.generateContainers(volumeMounts, containersPorts);
+    const podSpec = this.generatePodSpec(containers, volumes);
+
+    return podSpec;
```
wirednkod, line 25:
```suggestion
    return this.generatePodSpec(containers, volumes);
```

javascript/packages/orchestrator/src/providers/podman/resources/grafana.resource.ts
```diff
@@ -0,0 +1,113 @@
...
+      name: "grafana",
+      imagePullPolicy: "Always",
+      ports,
+      volumeMounts,
+    };
+
+    return [grafanaContainerSpec];
+  }
+
+  private generatePodSpec(containers: any[], volumes: any[]) {
```
wirednkod, line 88: same comment concerning `any` as before

javascript/packages/orchestrator/src/providers/podman/resources/introspector.resource.ts
```diff
@@ -0,0 +1,65 @@
+import { getRandomPort } from "@zombienet/utils";
+import { INTROSPECTOR_POD_NAME } from "../../../constants";
+
+export class IntrospectorResource {
+  constructor(
+    private readonly namespace: string,
+    private readonly wsUri: string,
+  ) {}
```
wirednkod, line 8: is there a need of constructor here?

javascript/packages/orchestrator/src/providers/podman/resources/introspector.resource.ts
```diff
@@ -0,0 +1,65 @@
...
+
+  public async generateSpec() {
+    const containerPorts = await this.generateContainersPorts();
+    const containers = this.generateContainers(containerPorts);
+    const podSpec = this.generatePodSpec(containers);
+
+    return podSpec;
```
wirednkod, line 15:
```suggestion
    return this.generatePodSpec(containers);
```

javascript/packages/orchestrator/src/providers/podman/resources/introspector.resource.ts
```diff
@@ -0,0 +1,65 @@
...
+      args: ["block-time-monitor", `--ws=${this.wsUri}`, "prometheus"],
+      imagePullPolicy: "Always",
+      ports,
+      volumeMounts: [],
+    };
+
+    return [introspectorContainerSpec];
+  }
+
+  private generatePodSpec(containers: any[]) {
```
wirednkod, line 41: same comment concerning `any` as before

javascript/packages/orchestrator/src/providers/podman/resources/introspector.resource.ts
```diff
@@ -0,0 +1,65 @@
...
+    const introspectorContainerSpec = {
+      image: "docker.io/paritytech/polkadot-introspector:latest",
+      name: INTROSPECTOR_POD_NAME,
+      args: ["block-time-monitor", `--ws=${this.wsUri}`, "prometheus"],
+      imagePullPolicy: "Always",
+      ports,
+      volumeMounts: [],
+    };
+
+    return [introspectorContainerSpec];
```
wirednkod, line 38:
```suggestion
    return [{
      image: "docker.io/paritytech/polkadot-introspector:latest",
      name: INTROSPECTOR_POD_NAME,
      args: ["block-time-monitor", `--ws=${this.wsUri}`, "prometheus"],
      imagePullPolicy: "Always",
      ports,
      volumeMounts: [],
    }];
```

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
```diff
@@ -0,0 +1,177 @@
...
+
+  constructor(
+    client: Client,
+    protected readonly namespace: string,
+    protected readonly nodeSetupConfig: Node,
+  ) {
+    const nodeName = nodeSetupConfig.name;
+    this.configPath = `${client.tmpDir}/${nodeName}/cfg`;
+    this.dataPath = `${client.tmpDir}/${nodeName}/data`;
+    this.relayDataPath = `${client.tmpDir}/${nodeName}/relay-data`;
```
wirednkod, line 25:
```suggestion
    const path = `${client.tmpDir}/${nodeSetupConfig.name}`;
    this.configPath = `${path}/cfg`;
    this.dataPath = `${path}/data`;
    this.relayDataPath = `${path}/relay-data`;
```

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
```diff
@@ -0,0 +1,177 @@
...
+    const volumes = await this.generateVolumes();
+    const volumeMounts = this.generateVolumesMounts();
+    const containersPorts = await this.generateContainersPorts();
+    const containers = await this.generateContainers(
+      volumeMounts,
+      containersPorts,
+    );
+    const podManifest = this.generatePodSpec(containers, volumes);
+
+    return podManifest;
```
wirednkod, line 38:
```suggestion
    return this.generatePodSpec(containers, volumes);
```

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
```diff
@@ -0,0 +1,177 @@
...
+    const dataVolume = {
+      name: "tmp-data",
+      hostPath: { type: "Directory", path: this.dataPath },
+    };
+    const relayDataVolume = {
+      name: "tmp-relay-data",
+      hostPath: { type: "Directory", path: this.relayDataPath },
+    };
+
+    return [configVolume, dataVolume, relayDataVolume];
```
wirednkod, line 63: Sorry for repeating that - IMO its better to avoid creating new vars if they are not reusable - just do the return at once; You can repeat wherever needed
```suggestion
    return [{
      name: "tmp-cfg",
      hostPath: { type: "Directory", path: this.configPath },
    }, {
      name: "tmp-data",
      hostPath: { type: "Directory", path: this.dataPath },
    }, {
      name: "tmp-relay-data",
      hostPath: { type: "Directory", path: this.relayDataPath },
    };];
```

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
```diff
@@ -0,0 +1,177 @@
...
+  }
+
+  private async portFromNodeSetupConfigOrDefault(
+    portProperty: keyof NonNullable<Node["externalPorts"]>,
+  ) {
+    const { externalPorts } = this.nodeSetupConfig;
+
+    if (externalPorts && portProperty in externalPorts) {
+      return externalPorts[portProperty];
+    }
```
wirednkod, line 93: Im not sure if what I am proposing here is correct but Isn't this the same?
```suggestion
    if (externalPorts?.portProperty)
      return externalPorts[portProperty];
```

wirednkod (commented):
Overall looks good - some suggestions on the writing only;

pepoviola (commented):
Hi @l0r1s, overall looks good! and is a great step forward in the refactor task.
I like the approach you take, and there are some areas of improving like error handling and logging. 

I'm not 100% convinced of the naming convention (with the `.resource`) but is something we can talk can agree with the rest of the team.

Thanks!!

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
l0r1s: You are right, it was a remainder when I was trying to find the correct type for the function argument

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
l0r1s: Yes, I'm use to it to make code cleaner with a quick glimpse, question of taste en colors, gonna fix it

javascript/packages/orchestrator/src/providers/podman/resources/node.resource.ts
wirednkod: I mean its just a proposal . if you think its a good idea go for it...same goes for most of the comments

PR #879, opened 2023-04-02
Refactor orchestrator

pepoviola:
- Small refactor, mostly moving things out from the orchestrator to `helpers`.
- Add relayChainCustomization fn that `drives` the customization of the spec instead of the orchestrator. I think this open the door to remove the read/write of each inner fn and just passing the spec as json.
- Small cleanup of `any` usage.


javascript/packages/orchestrator/src/chain-spec.ts
```diff
@@ -580,6 +580,57 @@ export async function getChainIdFromSpec(specPath: string): Promise<string> {
   });
 }
 
+export async function customizePlainRelayChain(
+  specPath: string,
+  networkSpec: ComputedNetwork,
+): Promise<void> {
+  // Relay-chain spec customization logic
```
wirednkod, line 587: since we are refactoring things - should we add also a try/catch in these functions?

javascript/packages/orchestrator/src/network-helpers/tracing-collator.ts
```diff
@@ -0,0 +1,79 @@
...
+              `service/${serviceName}`,
+              serviceNamespace,
+            );
+            network.tracing_collator_url = `http://localhost:${tracingPort}`;
+          } catch (_) {
+            console.log(
+              decorators.yellow(
+                `\n\t Warn: Can not create the forwarding to the tracing collator`,
+              ),
+            );
```
wirednkod, line 66: I think that we should always log the errors:

```suggestion
          } catch (err) {
            console.log(
              decorators.yellow(
                `\n\t Warn: Can not create the forwarding to the tracing collator`,
              )
            );
            console.error(err);
```

javascript/packages/orchestrator/src/providers/native/nativeClient.ts
```diff
@@ -431,6 +436,10 @@ export class NativeClient extends Client {
     return false;
   }
 
+  async spawnIntrospector(wsUri: string) {
+    // NOOP
+  }
```
wirednkod, line 441:
```suggestion
  async spawnIntrospector(wsUri: string) {
    // NOOP
    console.log("Not implemented");
  }
```

javascript/packages/orchestrator/src/spawner.ts
```diff
@@ -0,0 +1,249 @@
...
+    true,
+    node.p2pCertHash,
+  );
+
+  // multiAddressByNode[podDef.metadata.name] = nodeMultiAddress;
+
+  // if (node.addToBootnodes) {
+  //   bootnodes.push(nodeMultiAddress);
+  //   await addBootNodes(chainSpecFullPath, bootnodes);
+  // }
```
wirednkod, line 120: remove comments

wirednkod (approved):
lgtm - besides some small mentions

javascript/packages/orchestrator/src/chain-spec.ts
pepoviola: Sounds good, I will add now.

javascript/packages/orchestrator/src/providers/native/nativeClient.ts
pepoviola: I think is better to show a warning here, with the legend
```
We will not spawn an introspector instance on native provider.
```

javascript/packages/orchestrator/src/spawner.ts
pepoviola: Thanks, I think also I have to revisit the way bootnodes are handled.

javascript/packages/orchestrator/src/chain-spec.ts
```diff
@@ -580,6 +580,57 @@ export async function getChainIdFromSpec(specPath: string): Promise<string> {
...
+  const plainRelayChainSpec = readAndParseChainSpec(specPath);
+  const keyType = specHaveSessionsKeys(plainRelayChainSpec)
+    ? "session"
+    : "aura";
```
l0r1s, line 591: Could be good to have an enum for this, maybe create a issue like the ZombieRole one ?

javascript/packages/orchestrator/src/network-helpers/tracing-collator.ts
```diff
@@ -0,0 +1,80 @@
...
+    tracing_collator_url,
+    tracing_collator_service_port,
+    tracing_collator_service_name,
+    tracing_collator_service_namespace,
+  } = networkSpec.settings;
+  if (tracing_collator_url) network.tracing_collator_url = tracing_collator_url;
+  else {
+    const servicePort = tracing_collator_service_port || TRACING_COLLATOR_PORT;
+    switch (client.providerName) {
+      case "kubernetes":
```
l0r1s, line 28: Same, an enum for the provider type ?

javascript/packages/orchestrator/src/network-helpers/tracing-collator.ts
```diff
@@ -0,0 +1,80 @@
...
+          } catch (err) {
+            console.log(
+              decorators.yellow(
+                `\n\t Warn: Can not create the forwarding to the tracing collator`,
+              )
+            );
+            console.error(err);
+          }
+        }
+        break;
```
l0r1s, line 70: Maybe we should abstract this in the provider and just call a method for each provider ? What do you think ?

l0r1s: Because we are using a generic Client but at the same time checking for provider time which make the abstraction a less useful

javascript/packages/orchestrator/src/network.ts
```diff
@@ -95,16 +95,23 @@ export class Network {
...
+    namespace: string,
+    tmpDir: string,
+    startTime: number = new Date().getTime(),
+  ) {
     this.client = client;
     this.namespace = namespace;
     this.launched = false;
     this.wasRunning = false;
     this.tmpDir = tmpDir;
+    this.networkStartTime = startTime;
```
l0r1s, line 114:
```suggestion
  constructor(
    private readonly client: Client,
    private readonly namespace: string,
    private readonly tmpDir: string,
    private readonly networkStartTime: number = new Date().getTime(),
  ) {
```

javascript/packages/orchestrator/src/network.ts
```diff
@@ -95,16 +95,23 @@ export class Network {
   wasRunning: boolean;
```
l0r1s, line 95:
```suggestion
  wasRunning: boolean = false;
```

Can do the same for launched above

javascript/packages/orchestrator/src/network.ts
```diff
@@ -95,16 +95,23 @@ export class Network {
...
   tmpDir: string;
   backchannelUri: string = "";
+  chainId?: string;
   chainSpecFullPath?: string;
   tracing_collator_url?: string;
   networkStartTime?: number;
```
l0r1s, line 101:
```suggestion
  backchannelUri: string = "";
  chainId?: string;
  chainSpecFullPath?: string;
  tracing_collator_url?: string;
```

Can do the same with ```namespace``` and ```client``` above

javascript/packages/orchestrator/src/providers/client.ts
```diff
@@ -56,6 +56,7 @@ export abstract class Client {
   abstract startPortForwarding(
     port: number,
     identifier: string,
+    namespace?: string,
```
l0r1s, line 59: If it's available on the class, maybe not pass it as parameter ?

javascript/packages/orchestrator/src/spawner.ts
```diff
@@ -0,0 +1,249 @@
...
+    switch (client.providerName) {
+      case "podman":
+        logCommand = `podman logs -f ${podDef.metadata.name}_pod-${podDef.metadata.name}`;
+        break;
+      case "kubernetes":
+        logCommand = `kubectl logs -f ${podDef.metadata.name} -c ${podDef.metadata.name} -n ${namespace}`;
+        break;
+      case "native":
+        logCommand = `tail -f  ${client.tmpDir}/${podDef.metadata.name}.log`;
+        break;
```
l0r1s, line 242: Could probably be abstracted per provider

l0r1s (changes requested):
Some small changes to be made, maybe a rebase is needed to take the file naming branch that has been merged which creates conflicts

javascript/packages/orchestrator/src/network.ts
wirednkod:
```suggestion
  wasRunning = false;
```
The type is trivially inferred from the boolean literal thus the type annotation can be removed

javascript/packages/orchestrator/src/network.ts
wirednkod: same as above; also your proposal is removing some variables - why?
```suggestion
  tmpDir: string;
  backchannelUri = "";
  chainId?: string;
  chainSpecFullPath?: string;
  tracing_collator_url?: string;
  networkStartTime?: number;
```

javascript/packages/orchestrator/src/network.ts
wirednkod: In case we proceed with this, the `readonly` values need to be removed from the declaration of the class

javascript/packages/orchestrator/src/network.ts
l0r1s: Yes, because we are using constructor property definition below, we can remove them from here

javascript/packages/orchestrator/src/network.ts
l0r1s: It's a question of preferent but imho, it removed the boring part of plugging all property in the property definition so we have 2x less code

l0r1s (approved):

javascript/packages/orchestrator/src/network.ts
wirednkod: Could not agree more - I do not mind any of the 2 (and since ts -> CommonJS it does not really matter the output size); All I am saying is that we should be consistent across repo

pepoviola: I'm agree that we should stick to one option, at the moment we weren't using `readonly` properties in class constructors. But I see the benefit to switch to use it. I think we can address that in a separate `pr` if you both are agree?

l0r1s: Well, in fact I use readonly because most of the time my properties are shared non-mutable state but if you want to define a property mutable, you just remove the readonly (it's a mechanical habit kind of ahah), only putting a modifier like private/public/protected define the property automatically on the class, readonly is a nice-to-have to ensure no modification of non-mutable state

javascript/packages/orchestrator/src/providers/client.ts
pepoviola: There is a case where we use another `namespace`, from tracing services. That why is allowed to pass the namespace. I will leave for now, but we should clean that in the sdk.
Thanks!!

javascript/packages/orchestrator/src/spawner.ts
pepoviola: Done, thanks!

javascript/packages/orchestrator/src/chain-spec.ts
pepoviola: done, thanks!

PR #889, opened 2023-04-03
refactor(cli): split cli in actions files

pepoviola:
Split the `cli` in diff `action`s files.

### DIFF
--- a/javascript/packages/cli/src/actions/convert.ts
+++ b/javascript/packages/cli/src/actions/convert.ts
@@ -32,16 +32,13 @@ async function readInputFile(
   ext: string,
   fPath: string,
 ): Promise<PL_ConfigType> {
-  let json: object;
   if (ext === "json" || ext === "js") {
-    json =
-      ext === "json"
+     return  ext === "json"
         ? JSON.parse(fs.readFileSync(`${fPath}`, "utf8"))
         : await import(path.resolve(fPath));
-  } else {
-    throw Error("No valid extension was found.");
   }
-  return json;
+  
+  throw Error("No valid extension was found.");
 }
 
 async function convertInput(filePath: string) {
--- a/javascript/packages/cli/src/actions/convert.ts
+++ b/javascript/packages/cli/src/actions/convert.ts
@@ -7,7 +7,7 @@ import type {
 import { decorators, getFilePathNameExt } from "@zombienet/utils";
 import fs from "fs";
 import path from "path";
-import { DEFAULT_BALANCE } from "src/constants";
+import { DEFAULT_BALANCE } from "../constants";
 
 export async function convert(param: string) {
   try {
--- a/javascript/packages/cli/src/actions/convert.ts
+++ b/javascript/packages/cli/src/actions/convert.ts
@@ -1,0 +1,130 @@
+import type {
+  NodeConfig,
+  PL_ConfigType,
+  ParachainConfig,
+  PolkadotLaunchConfig,
+} from "@zombienet/orchestrator";
+import { decorators, getFilePathNameExt } from "@zombienet/utils";
+import fs from "fs";
+import path from "path";
+import { DEFAULT_BALANCE } from "src/constants";
+
+export async function convert(param: string) {
+  try {
+    const filePath = param;
+
+    if (!filePath) {
+      throw Error("Path of configuration file was not provided");
+    }
+
+    // Read through the JSON and write to stream sample
+    await convertInput(filePath);
+  } catch (err) {
+    console.log(
+      `\n ${decorators.red("Error: ")} \t ${decorators.bright(err)}\n`,
+    );
+  }
+}
+
+// Convert functions
+// Read the input file
+async function readInputFile(
+  ext: string,
+  fPath: string,
+): Promise<PL_ConfigType> {
+  let json: object;
+  if (ext === "json" || ext === "js") {
+    json =
+      ext === "json"
+        ? JSON.parse(fs.readFileSync(`${fPath}`, "utf8"))
+        : await import(path.resolve(fPath));
+  } else {
+    throw Error("No valid extension was found.");
+  }
+  return json;
+}
+
+async function convertInput(filePath: string) {
+  const { fullPath, fileName, extension } = getFilePathNameExt(filePath);
+
+  const convertedJson = await readInputFile(extension, filePath);
+
+  const { relaychain, parachains, simpleParachains, hrmpChannels, types } =
+    convertedJson;
+
+  let jsonOutput: PolkadotLaunchConfig;
+  const nodes: NodeConfig[] = [];
+  const paras: ParachainConfig[] = [];
+  let collators: NodeConfig[] = [];
+
+  const DEFAULT_NODE_VALUES = {
+    validator: true,
+    invulnerable: true,
+    balance: DEFAULT_BALANCE,
+  };
+
+  parachains &&
+    parachains.forEach((parachain: any) => {
+      collators = [];
+      parachain.nodes.forEach((n: any) => {
+        collators.push({
+          name: n.name,
+          command: "adder-collator",
+          ...DEFAULT_NODE_VALUES,
+        });
+      });
+      paras.push({
+        id: parachain.id,
+        collators,
+      });
+    });
+
+  collators = [];
+
+  simpleParachains &&
+    simpleParachains.forEach((sp: any) => {
+      collators.push({
+        name: sp.name,
+        command: "adder-collator",
+        ...DEFAULT_NODE_VALUES,
+      });
+      paras.push({
+        id: sp.id,
+        collators,
+      });
+    });
+
+  if (relaychain?.nodes) {
+    relaychain.nodes.forEach((n: any) => {
+      nodes.push({
+        name: `"${n.name}"`,
+        ...DEFAULT_NODE_VALUES,
+      });
+    });
+  }
+
+  jsonOutput = {
+    relaychain: {
+      default_image: "docker.io/paritypr/polkadot-debug:master",
+      default_command: "polkadot",
+      default_args: ["-lparachain=debug"],
+      chain: relaychain?.chain || "",
+      nodes,
+      genesis: relaychain?.genesis,
+    },
+    types,
+    hrmp_channels: hrmpChannels || [],
+    parachains: paras,
+  };
+
+  fs.writeFile(
+    `${fullPath}/${fileName}-zombienet.json`,
+    JSON.stringify(jsonOutput),
+    (error: any) => {
+      if (error) throw error;
+    },
+  );
+  console.log(
+    `Converted JSON config exists now under: ${fullPath}/${fileName}-zombienet.json`,
+  );
+}
--- a/javascript/packages/cli/src/actions/convert.ts
+++ b/javascript/packages/cli/src/actions/convert.ts
@@ -75,7 +75,6 @@ async function convertInput(filePath: string) {
       });
     });
 
-  collators = [];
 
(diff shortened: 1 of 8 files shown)


javascript/packages/cli/src/cli.ts
```diff
@@ -416,264 +148,21 @@ program
...
-    force,
-    inCI,
-    silent: false,
+function asyncAction(cmd: Function) {
+  return function () {
+    const args = [...arguments];
+    (async () => {
+      try {
+        if (cmd.name == "spawn") network = await cmd.apply(null, args);
+        else await cmd.apply(null, args);
```
l0r1s, line 159:
```suggestion
        if (cmd.name == "spawn") {
             network = await cmd.apply(null, args);
        }  else {
            await cmd.apply(null, args);
        }
```
Avoiding single line if is more readable

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+  let json: object;
+  if (ext === "json" || ext === "js") {
+    json =
+      ext === "json"
+        ? JSON.parse(fs.readFileSync(`${fPath}`, "utf8"))
+        : await import(path.resolve(fPath));
+  } else {
+    throw Error("No valid extension was found.");
+  }
+  return json;
```
l0r1s, line 44:
```suggestion
async function readInputFile(
  ext: string,
  fPath: string,
): Promise<PL_ConfigType> {
  if (ext === "json" || ext === "js") {
     return  ext === "json"
        ? JSON.parse(fs.readFileSync(`${fPath}`, "utf8"))
        : await import(path.resolve(fPath));
  }
  
  throw Error("No valid extension was found.");
```

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+          name: n.name,
+          command: "adder-collator",
+          ...DEFAULT_NODE_VALUES,
+        });
+      });
+      paras.push({
+        id: parachain.id,
+        collators,
+      });
+    });
```
l0r1s, line 80:
```suggestion
parachains &&
    parachains.forEach((parachain: any) => {
      collators = [];
      parachain.nodes.forEach((n: any) => {
        collators.push({
          name: n.name,
          command: "adder-collator",
          ...DEFAULT_NODE_VALUES,
        });
      });
      paras.push({
        id: parachain.id,
        collators,
      });
    });
```

What do you think, using destructuring and functionnal programming ?

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+  return json;
+}
+
+async function convertInput(filePath: string) {
+  const { fullPath, fileName, extension } = getFilePathNameExt(filePath);
+
+  const convertedJson = await readInputFile(extension, filePath);
+
+  const { relaychain, parachains, simpleParachains, hrmpChannels, types } =
+    convertedJson;
```
l0r1s, line 53:
```suggestion
  const { relaychain, parachains = [], simpleParachains = [], hrmpChannels, types } =
    convertedJson;
```

Default value so we can remove the condition checks

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+      collators.push({
+        name: sp.name,
+        command: "adder-collator",
+        ...DEFAULT_NODE_VALUES,
+      });
+      paras.push({
+        id: sp.id,
+        collators,
+      });
+    });
```
l0r1s, line 95:
```suggestion
  paras = paras.concat(
    simpleParachains.map(({ id, name }) => ({
      id,
      collators: [{ name, command: "adder-collator", ...DEFAULT_NODE_VALUES }],
    })),
  );
```
Destructuring and functionnal programming part 2

l0r1s: My question here is, because we don't reset the collators array in the old code, do we really want to have a parachain B to have the previous collator of parachain B ?

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+  const paras: ParachainConfig[] = [];
+  let collators: NodeConfig[] = [];
+
+  const DEFAULT_NODE_VALUES = {
+    validator: true,
+    invulnerable: true,
+    balance: DEFAULT_BALANCE,
+  };
+
+  parachains &&
```
l0r1s, line 66: ```suggestion```

Not needed, we have default in the destructuring at the start of the function

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+          ...DEFAULT_NODE_VALUES,
+        });
+      });
+      paras.push({
+        id: parachain.id,
+        collators,
+      });
+    });
+
+  collators = [];
```
l0r1s, line 82: Not needed if we use FP / destructuring

```suggestion
```

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  simpleParachains &&
```
l0r1s, line 84: Same here, we have default

```suggestion
```

javascript/packages/cli/src/actions/convert.ts
```diff
@@ -0,0 +1,130 @@
...
+    parachains: paras,
+  };
+
+  fs.writeFile(
+    `${fullPath}/${fileName}-zombienet.json`,
+    JSON.stringify(jsonOutput),
+    (error: any) => {
+      if (error) throw error;
+    },
+  );
```
l0r1s, line 126: If you use fs/promises instead of fs, you can simplify to:

```suggestion
  await fs.writeFile(
    `${fullPath}/${fileName}-zombienet.json`,
    JSON.stringify(jsonOutput)
  );
```

javascript/packages/cli/src/actions/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+
+  console.log(decorators.green("\n\n🧟🧟🧟 ZombieNet Setup 🧟🧟🧟\n\n"));
+  if (
+    ["aix", "freebsd", "openbsd", "sunos", "win32"].includes(process.platform)
+  ) {
+    console.log(
+      "Zombienet currently supports linux and MacOS. \n Alternative, you can use k8s or podman. For more read here: https://github.com/paritytech/zombienet#requirements-by-provider",
+    );
+    return;
+  }
```
l0r1s, line 28:
```suggestion
  if (!["linux", "darwin"].includes(process.platform)) {
    console.log(
      "Zombienet currently supports linux and MacOS. \n Alternative, you can use k8s or podman. For more read here: https://github.com/paritytech/zombienet#requirements-by-provider",
    );
    return;
  }
```

javascript/packages/cli/src/actions/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+/**
+ * Setup - easily download latest artifacts and make them executablein order to use them with zombienet
+ * Read more here: https://paritytech.github.io/zombienet/cli/setup.html
+ * @param params binaries that willbe downloaded and set up. Possible values: `polkadot` `polkadot-parachain`
+ * @returns
+ */
+export async function setup(params: any) {
+  const POSSIBLE_BINARIES = ["polkadot", "polkadot-parachain"];
+
+  console.log(decorators.green("\n\n🧟🧟🧟 ZombieNet Setup 🧟🧟🧟\n\n"));
```
l0r1s, line 20: Have you heard about the [colors](https://www.npmjs.com/package/colors) library ? 

It could enable to simplify the code like this (it extends the String.constructor):
```suggestion
  console.log("\n\n🧟🧟🧟 ZombieNet Setup 🧟🧟🧟\n\n".green);
```

Simpler and better because we don't have to manage our own "colors" system in utils.

javascript/packages/cli/src/actions/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+      `${decorators.yellow(
+        "Note: ",
+      )} You are using MacOS. Please, clone the polkadot repo ` +
+        decorators.cyan("(https://github.com/paritytech/polkadot)") +
+        ` and run it locally.\n At the moment there is no polkadot binary for MacOs.\n\n`,
+    );
+    const index = params.indexOf("polkadot");
+    if (index !== -1) {
+      params.splice(index, 1);
+    }
```
l0r1s, line 70:
```suggestion
params = params.filter(param => param !== "polkadot")
```

javascript/packages/cli/src/actions/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+  if (params.length === 0) {
+    console.log(decorators.green("No binaries to download. Exiting..."));
+    return;
+  }
+  let count = 0;
+  console.log("Setup will start to download binaries:");
+  params.forEach((a: any) => {
+    if (!POSSIBLE_BINARIES.includes(a)) {
+      const index = params.indexOf(a);
+      index > -1 && params.splice(index, 1);
```
l0r1s, line 82:
```suggestion
params = params.filter(param => param !== a);
```

javascript/packages/cli/src/actions/convert.ts
pepoviola: No, we don't want. This is a bug and the suggested code is ok.
Thanks!

javascript/packages/cli/src/actions/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+    tag_name = release.tag_name;
+
+    if (!tag_name) {
+      throw new Error(
+        "Should never come to this point. Tag_name should never be undefined!",
+      );
+    }
+
+    return [
+      `https://github.com/paritytech/${repo}/releases/download/${tag_name}/${name}`,
```
l0r1s, line 220: The asset from Github expose the "browser_download_url" property so we can replace it with:
```suggestion
      obj.browser_download_url
```

Screenshot from 2023-04-05 15-39-59;

javascript/packages/cli/src/actions/setup.ts
```diff
@@ -0,0 +1,235 @@
...
+    } else if (err.response && err.response.status === 404) {
+      throw new Error(
+        "Could not find a release. Error 404 (not found) detected",
+      );
+    }
+    throw new Error(
+      `Error status: ${err?.response?.status}. Error message: ${err?.response}`,
+    );
+  }
+};
```
l0r1s, line 235: That's why I don't like axios and prefer [wretch](https://github.com/elbywan/wretch), the code can be simplified like this:

```ts
wretch("anything")
  .get()
  .notFound(error => { /* ... */ })
  .unauthorized(error => { /* ... */ })
  .error(418, error => { /* ... */ })
  .res(response => /* ... */)
  .catch(error => { /* uncaught errors */ })
  ```

javascript/packages/cli/src/actions/convert.ts
pepoviola: We should change the `import` to use fs.promises here.

l0r1s: Yes, the exact import is ```import {...} from "fs/promises```

javascript/packages/cli/src/actions/setup.ts
pepoviola: Yes, we are moving away from `axios`. I think @wirednkod is doing this in another pr.

PR #905, opened 2023-04-07
Minor Unit tests for `Utils` package for test env

wirednkod:
This PR sets the test environment, and introduces a pre-push hook for the tests to run.
Small amount of tests have been added inside the utils package;


pepoviola (approved):

javascript/.husky/pre-commit
```diff
@@ -1,4 +1,4 @@
 #!/usr/bin/env sh
 cd javascript
 
-npm run lint
+npm run lint && npm run test
```
pepoviola, line 4: I think we should run the `test` in pre-push hook. Allowing us to commit `wip`s
Thanks!

PR #922, opened 2023-04-11
Add pre-commit hook for linter

wirednkod:
This way - prior to a commit the linter will run and we will avoid any extra runs breaking in the CI'


pepoviola:
In `Ci` we don't install `devDeps` so this setup fails https://gitlab.parity.io/parity/mirrors/zombienet/-/jobs/2668470. We can workaround installing husky directly.

Thanks!

pepoviola (approved):

PR #933, opened 2023-04-14
[CI] Podman tests 

l0r1s:


pepoviola:
Looks good! Can we use the `artifacts` from the `build` step to not build here again?

.github/workflows/podman-ci-reusable.yml
```diff
@@ -0,0 +1,51 @@
...
+      - name: Run test
+        run: npm run zombie -- --provider podman test ${{ inputs.test_path }}
+        env:
+          DEBUG: zombie*
+          ZOMBIENET_INTEGRATION_TEST_IMAGE: docker.io/paritypr/polkadot-debug:master
+          COL_IMAGE: docker.io/paritypr/colander:master
+          MALUS_IMAGE: 
+          BIFROST_COL_IMAGE: docker.io/bifrostnetwork/bifrost:latest
+          MOONBEAM_COL_IMAGE: docker.io/purestake/moonbeam:v0.26
+          OAK_COL_IMAGE: docker.io/oaknetwork/turing:latest
```
pepoviola, line 50:
```suggestion
          OAK_COL_IMAGE: docker.io/oaknetwork/turing:latest
          RUN_IN_CONTAINER: 1
```
Just for trying .

.github/workflows/podman-ci-reusable.yml
```diff
@@ -0,0 +1,52 @@
...
+        run: npm run zombie -- --provider podman test ${{ inputs.test_path }}
+        env:
+          DEBUG: zombie*
+          ZOMBIENET_INTEGRATION_TEST_IMAGE: docker.io/paritypr/polkadot-debug:master
+          COL_IMAGE: docker.io/paritypr/colander:master
+          MALUS_IMAGE: docker.io/paritypr/malus:4131-ccd09bbf
+          BIFROST_COL_IMAGE: docker.io/bifrostnetwork/bifrost:latest
+          MOONBEAM_COL_IMAGE: docker.io/purestake/moonbeam:v0.26
+          OAK_COL_IMAGE: docker.io/oaknetwork/turing:latest
+          RUN_IN_CONTAINER: 1
```
pepoviola, line 51:
```suggestion
```

tests/smoke/0001-smoke.zndsl
```diff
@@ -0,0 +1,42 @@
+Description: Smoke Test
+Network: ./0001-smoke.toml
+Creds: config
+
+alice: is up
+bob: is up
+
+# Logs assertions
+alice: reports block height is at least 10 within 200 seconds
```
pepoviola, line 9:
```suggestion
alice: reports block height is at least 4 within 200 seconds
```

tests/smoke/0001-smoke.zndsl
```diff
@@ -0,0 +1,42 @@
...
+alice: is up
+alice: restart after 20 seconds
+sleep 25 seconds
+alice: is up
+alice: pause
+# Todo add `should fail` support
+# alice: is down
+alice: resume
+sleep 25 seconds
+alice: is up
```
pepoviola, line 42:
```suggestion
alice: is up between 30 seconds
```

tests/smoke/0001-smoke.zndsl
```diff
@@ -0,0 +1,42 @@
...
+alice: restart after 20 seconds
+sleep 25 seconds
+alice: is up
+alice: pause
+# Todo add `should fail` support
+# alice: is down
+alice: resume
+sleep 25 seconds
+alice: is up between 30 seconds
```
pepoviola, line 42:
```suggestion
alice: is up within 30 seconds
```

tests/smoke/0001-smoke.zndsl
```diff
@@ -0,0 +1,42 @@
+Description: Smoke Test
+Network: ./0001-smoke.toml
+Creds: config
+
+alice: is up
+bob: is up
+
+# Logs assertions
+alice: reports block height is at least 4 within 200 seconds
+alice: log line contains "Imported #12" within 20 seconds
```
pepoviola, line 10:
```suggestion
alice: log line contains "Imported #2" within 20 seconds
```

tests/smoke/0001-smoke.zndsl
```diff
@@ -0,0 +1,42 @@
...
+alice: log line matches glob "*rted #1*" within 10 seconds
+alice: log line matches "Imported #[0-9]+" within 10 seconds
+alice: count of log lines containing "imported" is at least 10 within 10 seconds
```
pepoviola, line 13:
```suggestion
alice: count of log lines containing "imported" is at least 4 within 10 seconds
```

pepoviola (approved):

.github/workflows/podman-ci-reusable.yml
```diff
@@ -0,0 +1,66 @@
+name: ZombieNet Reusable Podman CI
```
wirednkod, line 1: There should not be any reusable CIs

All of the install, lint, build, tests, provider tests etc should run in 1 CI for github for all the possible node versions

.github/workflows/podman-ci.yml
```diff
@@ -0,0 +1,20 @@
+name: ZombieNet Podman CI
```
wirednkod, line 1: same as above

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -97,15 +97,17 @@ export class NetworkNode implements NetworkNodeInterface {
...
+    const result = await client.runCommand(args, { scoped });
     return result.exitCode === 0;
   }
 
   async resume() {
     const client = getClient();
     const args = client.getResumeArgs(this.name);
-    const result = await client.runCommand(args, { scoped: true });
+    const scoped = client.providerName === "kubernetes";
+    const result = await client.runCommand(args, { scoped });
```
wirednkod, line 110: can this repeatable code (with the `const client = getClient();` become reusable in the class maybe?

javascript/packages/orchestrator/src/providers/podman/podmanClient.ts
```diff
@@ -489,20 +489,32 @@ export class PodmanClient extends Client {
...
+      `${name}_pod-${name}`,
+      "bash",
+      "-c",
+      "echo resume > /tmp/zombiepipe",
+    ];
   }
 
   async restartNode(name: string, timeout: number | null): Promise<boolean> {
-    const args = ["exec", name, "--", "bash", "-c"];
+    const args = ["exec", `${name}_pod-${name}`, "bash", "-c"];
```
wirednkod, line 511: why this change? Why not:
```suggestion
    const args = ["exec", `pod_${name}`, "bash", "-c"];
```

javascript/packages/orchestrator/src/providers/podman/resources/tempoResource.ts
```diff
@@ -38,7 +38,7 @@ export class TempoResource {
         __dirname,
         `./configs/tempo.yaml`,
       );
-      await fs.writeFile(`${this.configPath}/tempo.yaml`, templateConfigPath);
+      await fs.copyFile(templateConfigPath, `${this.configPath}/tempo.yaml`);
```
wirednkod, line 41: why is this change needed?

javascript/packages/orchestrator/src/providers/podman/resources/prometheusResource.ts
```diff
@@ -44,9 +44,9 @@ export class PrometheusResource {
         __dirname,
         "./configs/prometheus.yml",
       );
-      await fs.writeFile(
-        `${this.configPath}/prometheus.yml`,
+      await fs.copyFile(
```
wirednkod, line 47: why is this change needed?

javascript/packages/orchestrator/src/providers/podman/resources/configs/tempo.yaml
```diff
@@ -33,11 +33,11 @@ storage:
     backend: local                     # backend configuration to use
     block:
       bloom_filter_false_positive: .05 # bloom filter false positive rate.  lower values create larger filters but fewer false positives
-      index_downsample_bytes: 1000     # number of bytes per index record
-      encoding: zstd                   # block encoding/compression.  options: none, gzip, lz4-64k, lz4-256k, lz4-1M, lz4, snappy, zstd, s2
+      v2_index_downsample_bytes: 1000     # number of bytes per index record
```
wirednkod, line 36: why are these change needed?

tests/smoke/0001-custom.sh
```diff
@@ -0,0 +1,2 @@
+# just echo for this test
+echo "OK"
```
wirednkod, line 2: do we need this extra file for just an echo?

wirednkod (changes requested):

javascript/packages/orchestrator/src/networkNode.ts
pepoviola: At the moment, `client` is a `singleton` we can refactor this to have get the client in the class constructor and then use `this.client`. But I think is not in the scope of this `pr`.

javascript/packages/orchestrator/src/providers/podman/podmanClient.ts
pepoviola: Podman, at least in the version we have support, use `name_pod-name` as container identifier and that why we should use that way. I removed the `"--"` because that argument calls doesn't work in podman is from `k8s`. Since we never tested this particular fn in podman because should never work the were was it.

tests/smoke/0001-custom.sh
pepoviola: Yes, we need a `bash` script to test the `bash` assertion. Could be empty also, but I think have just an `echo` is better.

javascript/packages/orchestrator/src/providers/podman/resources/tempoResource.ts
l0r1s: Because writeFile doesn't works that way, it takes the content of the file, while copyFile take the path of the file

javascript/packages/orchestrator/src/providers/podman/resources/prometheusResource.ts
l0r1s: Same as above

.github/workflows/podman-ci-reusable.yml
l0r1s: What do you mean ? Reusable workflow are like a function, this is for code reuse and avoid duplication, because we want to run both tests in parallel

pepoviola (approved):

pepoviola:
Ping @wirednkod, can you review this one again. I think is good to merge.
Thanks!

wirednkod (commented):
> Ping @wirednkod, can you review this one again. I think is good to merge. Thanks!

I mean - I see still things that I commented not being altered - I still think that the Provider's tests should run in one CI and not a separate one; In addition I am not for, on caching deps;
There is a repeating of install/build, actions that take place in the existing CI - this will just make things duplicated;

wirednkod:
I mean we can just do something like this:

```yml
name: ZombieNet Basic CI

on:
  push:
    branches: [main]
  pull_request:
    branches: [main]

jobs:
  build:
    runs-on: ubuntu-latest

    strategy:
      matrix:
        node-version: [18.x, 19.x]

    steps:
      - uses: actions/checkout@v3.5.2
      - name: Use Node.js ${{ matrix.node-version }}
        uses: actions/setup-node@v3
        with:
          node-version: ${{ matrix.node-version }}
      - run: npm install
        working-directory: "./javascript"
      - run: npm run build
        working-directory: "./javascript"
      - run: npm run lint
        working-directory: "./javascript"
      - run: npm run test
        working-directory: "./javascript"

      - name: Upload build artifacts
        uses: actions/upload-artifact@v3
        with:
          name: ${{ runner.os }}-build-${{ github.sha }}
          path: |
            javascript/packages/cli/dist
            javascript/packages/orchestrator/dist
            javascript/packages/utils/dist

  build-rust:
    runs-on: ubuntu-latest

    steps:
      - uses: actions/checkout@v3.5.2
      - name: Install Rust Stable toolchain
        uses: actions-rs/toolchain@v1
        with:
          profile: minimal
          toolchain: stable
          override: true

      - name: Run cargo fmt
        run: cargo fmt --all -- --check

      - uses: actions-rs/clippy-check@v1
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
          args: --all-features

      - name: Build
        run: cargo build

      - name: Run tests
        run: cargo test --verbose --all-targets --all-features

  run-podman-tests:
    runs-on: ubuntu-22.04
    needs: build-artifacts
    strategy:
      matrix:
        test-path:
          - ../tests/smoke/0001-smoke.zndsl
          - ../tests/0013-db-snapshot.zndsl

    steps:
      - name: Checkout repository
        uses: actions/checkout@v3.5.0

      - name: Setup containers registries
        run: |
          mkdir -p /etc/containers \
          && echo "unqualified-search-registries = ['docker.io']" \
          && sudo tee /etc/containers/registries.conf
      - name: Run test
        run: npm run zombie -- --provider podman test ${{ matrix.test-path }}
        working-directory: ./javascript
        env:
          DEBUG: zombie*
          ZOMBIENET_INTEGRATION_TEST_IMAGE: docker.io/paritypr/polkadot-debug:master
          COL_IMAGE: docker.io/paritypr/colander:master
          MALUS_IMAGE: docker.io/paritypr/malus:4131-ccd09bbf

  all:
    # This dummy job depends on all the mandatory checks. It succeeds if and only if all CI checks
    # are successful.
    needs: [build, build-rust, run-podman-tests]
    runs-on: ubuntu-latest
    steps:
      - run: echo Success
```

There is no reason for podman tetsts to run e.g. if rust build fail, or some other stop of the CI fails..

pepoviola:
Yes, sounds good to me. Thanks @wirednkod / @louismerlin :)

.github/workflows/ci.yml
```diff
@@ -55,10 +65,54 @@ jobs:
...
+      - name: Setup containers registries
+        run: |
+          mkdir -p /etc/containers \
+          && echo "unqualified-search-registries = ['docker.io']" \
+          && sudo tee /etc/containers/registries.conf
+
+      - name: Setup NodeJS
+        uses: actions/setup-node@v3
+        with:
+          node-version: 18.x
```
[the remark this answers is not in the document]
pepoviola: I think is ok to only test on `18`, we don't need to run the same integration test on both 18 and 19.

.github/workflows/ci.yml
```diff
@@ -55,10 +65,54 @@ jobs:
...
+
+      - name: Retrieve build artifacts
+        uses: actions/download-artifact@master
+        with:
+          name: ${{ runner.os }}-build-${{ github.sha }}
+          path: ./javascript/packages
```
wirednkod, line 97: I think this should be removed as well

.github/workflows/ci.yml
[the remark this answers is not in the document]
wirednkod: Mentioned in another comment that these lines are actually not needed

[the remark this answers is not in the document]
l0r1s: Well, I think this is still needed to pin the version of node being used instead of having a lts used and updated when we don't know, no ?

.github/workflows/ci.yml
```diff
@@ -29,6 +29,16 @@ jobs:
       - run: npm run test
         working-directory: "./javascript"
 
+      - name: Upload build artifacts
+        if: ${{ matrix.node-version == '18.x' }}
```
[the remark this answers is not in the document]
l0r1s: So we don't cache anymore the build and build 2 times ?

.github/workflows/ci.yml
l0r1s: Well, If I remove this, we can't use node in the job because it's started on a new runner (not a step)

wirednkod: Why you need to retrieve build artifacts since we want to remove also the upload?

wirednkod (approved):

Issue #901, opened 2023-04-05
Refacto: zombieRole interface to enum


l0r1s:
The zombieRole is a interface and is present in multiple location in the codebase. We should make an enum for this and refactor where it is being used.

PR #938, opened 2023-04-17
chore: changed raw string zombie role to enum

l0r1s:
Refacto for https://github.com/paritytech/zombienet/issues/901


pepoviola (approved):
Looks good! thanks, I think there are a couple of formating issues.

wirednkod (approved):

wirednkod:
> Looks good! thanks, I think there are a couple of formating issues.

Looks good. I wonder why the PR got to open, since we have the hooks now 🤔

l0r1s:
I added the missing "postinstall" script to make husky install automatic when you run "npm install"

l0r1s:
Fixed following husky doc to avoid running in pipeline

Issue #782, opened 2023-03-10
Document `convert` command


wirednkod:
Once #550 is done we need to add documentation for `convert`

PR #955, opened 2023-04-24
docs: added convert command to documentation

l0r1s:
Resolves https://github.com/paritytech/zombienet/issues/782


wirednkod (changes requested):
`Polkadot launch` is not the "old" project while `Zombienet` being the "new";
They are 2 different projects, with Polkadot Launch not being supported any more; As Parity we provide Zombienet as an alternative solution but not as the replacement;

wirednkod (approved):

PR #969, opened 2023-04-26
Add Timeout function in utils and bump version

wirednkod:


javascript/packages/utils/src/misc.ts
```diff
@@ -120,6 +120,12 @@ export function getLokiUrl(
   return loki_url;
 }
 
+export const Timeout = (time: number) => {
```
pepoviola, line 123: I think we could use a more concrete naming here, something like `TimeoutAbortController`. wdyt?

pepoviola (approved):

PR #975, opened 2023-04-28
feat(orchestrator): switch to concurrency 1 when network definiton us…

pepoviola:
…e references.

- When network definition use references (e.g `{{'dave'|zombie('wsUri')}}`) we need to switch to single concurrency to prevent race conditions.

Thanks!

### DIFF
--- a/javascript/packages/orchestrator/src/network.ts
+++ b/javascript/packages/orchestrator/src/network.ts
@@ -9,6 +9,7 @@ import {
   BAKCCHANNEL_PORT,
   BAKCCHANNEL_URI_PATTERN,
   DEFAULT_INDIVIDUAL_TEST_TIMEOUT,
+  TOKEN_PLACEHOLDER,
 } from "./constants";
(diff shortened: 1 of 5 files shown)


wirednkod (approved):

javascript/packages/orchestrator/src/network.ts
```diff
@@ -369,7 +370,8 @@ export class Network {
 
   replaceWithNetworInfo(placeholder: string): string {
     return placeholder.replace(
-      /{{ZOMBIE:(.*?):(.*?)}}/gi,
+      ///{{ZOMBIE:(.*?):(.*?)}}/gi,
```
l0r1s, line 373: Maybe removing this comment ?

l0r1s (commented):
lgtm, except the comment to be remove :+1:

pepoviola:
> lgtm, except the comment to be remove :+1: 

Ahh I just saw this. I will remove in a follow up pr

PR #977, opened 2023-04-28
fix dedup logs

### LINKED ISSUE #580: Multiple `line matches` in test do not work
Multiple `line matches` in test does not work. Only first `line match` works correctly, the second one times out.

It maybe reproduced with the  [`0001-small-network.zndsl`](https://github.com/paritytech/zombienet/blob/main/examples/0001-small-network.zndsl) example.
```
zombienet-linux test --provider podman 0001-small-network.zndsl
```

The error is:
```
┌──────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐                                                                                                          
│ 11/11/2022, 8:47:18 AM       │ ✅ alice: reports histogram polkadot_pvf_execution_time has at least 2 samples in buckets ["0.1",… │                                                                                                          
└──────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────┘                                                                                                          
┌──────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐                                                                                                          
│ 11/11/2022, 8:47:18 AM       │ ✅ bob: log line matches glob "*rted #1*" within 10 seconds (172ms)                                │                                                                                                          
└──────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────┘                                                                                                          

         Error:                                            
                 Timeout(10), "getting log pattern Imported #[0-9]+ within 10 secs".                                                                                                                                                           

┌──────────────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐                                                                                                          
│ 11/11/2022, 8:47:28 AM       │ ❌ bob: log line matches "Imported #[0-9]+" within 10 seconds (10162ms)                            │                                                                                                          
└──────────────────────────────┴────────────────────────────────────────────────────────────────────────────────────────────────────┘ 
```

`bob.log` contains `Imported #[0-9]` pattern:
```
$ cat /tmp/zombie-1753d204_-754983-5DF8mNV09fWZ/logs/bob.log | grep "Imported #"
2022-11-11 07:46:06.016  INFO tokio-runtime-worker substrate: ✨ Imported #1 (0xfb52…0f8f)    
2022-11-11 07:46:12.013  INFO tokio-runtime-worker substrate: ✨ Imported #2 (0x81d9…690c)    
2022-11-11 07:46:18.010  INFO tokio-runtime-worker substrate: ✨ Imported #3 (0xcb13…bbf7)    
2022-11-11 07:46:24.017  INFO tokio-runtime-worker substrate: ✨ Imported #4 (0x99de…540c)    
2022-11-11 07:46:30.016  INFO tokio-runtime-worker substrate: ✨ Imported #5 (0x372a…846d)   
```

pepoviola:
Fixes #580 

And our current CI.

### DIFF
--- a/javascript/packages/orchestrator/src/networkNode.ts
+++ b/javascript/packages/orchestrator/src/networkNode.ts
@@ -40,8 +40,6 @@ export class NetworkNode implements NetworkNodeInterface {
   userDefinedTypes: any;
   para?: PARA;
   parachainId?: number;
-  lastLogLineCheckedTimestamp?: string;
-  lastLogLineCheckedIndex?: number;
   group?: string;
 
   constructor(
@@ -415,6 +413,8 @@ export class NetworkNode implements NetworkNodeInterface {
     timeout: number = DEFAULT_INDIVIDUAL_TEST_TIMEOUT,
   ): Promise<boolean> {
     try {
+      let lastLogLineCheckedTimestamp: string;
+      let lastLogLineCheckedIndex: number;
       const re = isGlob ? makeRe(pattern) : new RegExp(pattern, "ig");
       if (!re) throw new Error(`Invalid glob pattern: ${pattern} `);
       const client = getClient();
@@ -425,7 +425,10 @@ export class NetworkNode implements NetworkNodeInterface {
           const dedupedLogs = this._dedupLogs(
             logs.split("\n"),
             client.providerName === "native",
+            lastLogLineCheckedTimestamp,
+            lastLogLineCheckedIndex
           );
+
           const index = dedupedLogs.findIndex((line) => {
             if (client.providerName !== "native") {
               // remove the extra timestamp
@@ -436,10 +439,10 @@ export class NetworkNode implements NetworkNodeInterface {
 
           if (index >= 0) {
             done = true;
-            this.lastLogLineCheckedTimestamp = dedupedLogs[index];
-            this.lastLogLineCheckedIndex = index;
+            lastLogLineCheckedTimestamp = dedupedLogs[index];
+            lastLogLineCheckedIndex = index;
             debug(
-              this.lastLogLineCheckedTimestamp.split(" ").slice(1).join(" "),
+              lastLogLineCheckedTimestamp.split(" ").slice(1).join(" "),
             );
           } else {
             await new Promise((resolve) => setTimeout(resolve, 1000));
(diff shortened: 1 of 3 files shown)


wirednkod (approved):

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -425,7 +425,10 @@ export class NetworkNode implements NetworkNodeInterface {
           const dedupedLogs = this._dedupLogs(
             logs.split("\n"),
             client.providerName === "native",
+            lastLogLineCheckedTimestamp,
+            lastLogLineCheckedIndex,
```
l0r1s, line 429: It won't cause issue if it's not initialized ?

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -538,12 +539,17 @@ export class NetworkNode implements NetworkNodeInterface {
...
-    const lastLineTs = this.lastLogLineCheckedTimestamp.split(" ")[0];
+  // prevent to search in the same log line twice.
+  _dedupLogs(
+    logs: string[],
+    useIndex = false,
+    lastLogLineCheckedTimestamp: string,
+    lastLogLineCheckedIndex: number,
+  ): string[] {
+    if (!lastLogLineCheckedTimestamp) return logs;
+    if (useIndex) return logs.slice(lastLogLineCheckedIndex);
```
l0r1s, line 550: Here, lastLogLineCheckedIndex is undefined when entering the above loop

l0r1s (commented):
Some questions about the variable initialization

javascript/packages/orchestrator/src/networkNode.ts
pepoviola: Well, those two variables work in tandem to represent the different providers features. The invariant here is that `lastLogLineCheckedIndex` will be undefined at the same time that `lastLogLineCheckedTimestamp`  is undefinded, so the this if (https://github.com/paritytech/zombienet/pull/977/files#diff-fc4624d6bb61da0712b154b5018f73681adaf8edadb4236a0b5293ea915ef41eR549) cover this case.

I think there is room to improving here, but we can do it in a follow up pr.

Thanks!

PR #983, opened 2023-05-01
Feat/k8s service

l0r1s:
* Updated pod ``RestartPolicy`` with k8s provider to ``Always``
* Added one service per pod and updated node identifier to match this and use service
* Fixed a bug with label on pod not using the actual zombie role


pepoviola:
Thanks @l0r1s, looks good. We should also update the WS_URI and PROMETHEUS_URI to the service name when is used in our `ci` (https://github.com/paritytech/zombienet/pull/983/files#diff-24b7795ec93992e92b59b879677a7a7e90adcdf07284b6910a165592b680dabfL124-L133).

Thanks!

l0r1s (commented):
:+1:

pepoviola (approved):

PR #1001, opened 2023-05-03
Add prometheus prefix solution

### LINKED ISSUE #974: Customizable metrics prefixes
### Is your feature request related to a problem? Please describe.

We were trying to integrate KAGOME into the zombienet, and it almost worked. The only missing thing is that zombienet expected prometheus metric with `substrate_` prefix, while in KAGOME we have `kagome_`. 
Eg **substrate**\_process_start_time_seconds instead of **kagome**\_process_start_time_seconds.

### Describe the solution you'd like

It would be convenient to customize prefix metric. This could be done as new parameter. For example:
```
  [[relaychain.nodes]]
  name = "bob"
  command = "./bins/kagome"
  prometheus_prefix="kagome"
```
Default value for `prometheus_prefix` could be `substrate` for backwards compatibility

### Describe alternatives you've considered

_No response_

### Additional context

KAGOME is a C++ implementation of Polkadot host. We have parachains support, live validator in westend. Would be great to add more tests into our CI using zombienet

https://github.com/soramitsu/kagome

wirednkod:
Fixes #974

### DIFF
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -129,6 +129,10 @@ export async function generateNetworkSpec(
     parachains: [],
   };
 
+  if (config.relaychain.prometheus_prefix)
+    networkSpec.relaychain.prometheusPrefix =
+      config.relaychain.prometheus_prefix;
+
   // check all imageURLs for validity
   // TODO: These checks should be agains all config items that needs check
   configurationFileChecks(config);
@@ -207,6 +211,12 @@ export async function generateNetworkSpec(
           nodeGroup.resources || networkSpec.relaychain.defaultResources,
         db_snapshot: nodeGroup.db_snapshot,
       };
+      if (nodeGroup.prometheus_prefix) {
+        node.prometheus_prefix = nodeGroup.prometheus_prefix;
+      } else if (config.relaychain.prometheus_prefix) {
+        node.prometheus_prefix = config.relaychain.prometheus_prefix;
+      }
+
       const nodeSetup = await getNodeFromConfig(
         networkSpec,
         node,
@@ -256,7 +266,7 @@ export async function generateNetworkSpec(
           await getCollatorNodeFromConfig(
             networkSpec,
             collatorConfig,
-            parachain.id,
+            parachain,
             paraChainName,
             para,
             bootnodes,
@@ -285,7 +295,7 @@ export async function generateNetworkSpec(
             await getCollatorNodeFromConfig(
               networkSpec,
               node,
-              parachain.id,
+              parachain,
               paraChainName,
               para,
               bootnodes,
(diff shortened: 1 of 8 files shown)


javascript/packages/orchestrator/src/configGenerator.ts
```diff
@@ -207,6 +211,12 @@ export async function generateNetworkSpec(
           nodeGroup.resources || networkSpec.relaychain.defaultResources,
         db_snapshot: nodeGroup.db_snapshot,
       };
+      if (nodeGroup.prometheus_prefix) {
+        node.prometheus_prefix = nodeGroup.prometheus_prefix;
+      } else if (config.relaychain.prometheus_prefix) {
+        node.prometheus_prefix = config.relaychain.prometheus_prefix;
+      }
+
```
l0r1s, line 219: How about this ?

```suggestion
        node.prometheus_prefix = nodeGroup.prometheus_prefix || config.relaychain.prometheus_prefix;
```

javascript/packages/orchestrator/src/configGenerator.ts
```diff
@@ -618,6 +631,9 @@ async function getNodeFromConfig(
     p2pCertHash: node.p2p_cert_hash,
   };
 
+  if (node.prometheus_prefix)
+    nodeSetup.prometheusPrefix = node.prometheus_prefix;
```
l0r1s, line 635: If ```prometheus_prefix``` is ```some value || undefined``` it will stay ```undefined``` 

```suggestion
nodeSetup.prometheusPrefix = node.prometheus_prefix;
```

javascript/packages/orchestrator/src/metrics/index.ts
```diff
@@ -127,8 +127,8 @@ export function getMetricName(metricName: string): string {
   return metricNameTouse;
 }
 
-export function getProcessStartTimeKey() {
-  return "substrate_process_start_time_seconds";
+export function getProcessStartTimeKey(prefix = "substrate") {
+  return `${prefix}_process_start_time_seconds`;
```
l0r1s, line 131: :+1:

l0r1s (approved):
Some contestable remarks but lgtm :+1:

javascript/packages/orchestrator/src/configGenerator.ts
wirednkod: not really; cause there is  a chance that none of them exist, and in that case I want to avoid adding the key "prometheus_prefix" to the object

javascript/packages/orchestrator/src/configGenerator.ts
wirednkod: Not really - prior to this if the key `prometheusPrefix` does not exist; With your suggestion, it will create the key and "assign" `undefined` to it

pepoviola:
thanks @wirednkod, looks good! I think we can take another approach by `set` the prefix (with `substrate` as default) if not set by config, this will save us some `if` statements down the road :)
If you are agree, I can make the modifications.

Thanks!

l0r1s (approved):

Issue #757, opened 2023-02-26
[bug] No way to teardown network with podman


pepoviola:
Since with `podman` provider zombienet is exiting after spawn and there is no easy way to teardown the network. Only with `podman rm ...` manually.

pepoviola:
Hi @l0r1s, did you send SIGINT before complete launch? because iirc the issue is that the process exit after the network is spawned instead of keep alive as does with `native/k8s`.

Thanks!

wirednkod:
@l0r1s not fixed yet - in my todo list

PR #1022, opened 2023-05-09
Bug/teardown podman

l0r1s:
Fix https://github.com/paritytech/zombienet/issues/757

* Fixed podman issue when trying to teardown before fully spawned.
* Fixed podman issue when the network is spawned to prevent exiting zombienet

### DIFF
--- a/javascript/packages/cli/src/actions/spawn.ts
+++ b/javascript/packages/cli/src/actions/spawn.ts
@@ -30,7 +30,8 @@ export async function spawn(
   credsFile: string | undefined,
   cmdOpts: any,
   program: any,
-): Promise<Network> {
+  setGlobalNetwork: (network: Network) => void,
+): Promise<void> {
   const opts = { ...program.parent.opts(), ...cmdOpts };
   const dir = opts.dir || "";
   const force = opts.force || false;
@@ -93,8 +94,8 @@ export async function spawn(
     force,
     inCI,
     silent: false,
+    setGlobalNetwork,
   };
   const network = await start(creds, config, options);
   network.showNetworkInfo(config.settings?.provider);
-  return network;
 }
--- a/javascript/packages/cli/src/actions/spawn.ts
+++ b/javascript/packages/cli/src/actions/spawn.ts
@@ -98,4 +98,6 @@ export async function spawn(
   };
   const network = await start(creds, config, options);
   network.showNetworkInfo(config.settings?.provider);
+  // keep the process running
+  await new Promise(() => setInterval(() => {}, 1000));
 }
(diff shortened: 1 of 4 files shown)


javascript/packages/cli/src/actions/spawn.ts
```diff
@@ -93,8 +94,10 @@ export async function spawn(
     force,
     inCI,
     silent: false,
+    setGlobalNetwork,
   };
   const network = await start(creds, config, options);
   network.showNetworkInfo(config.settings?.provider);
-  return network;
+  // keep the process running
+  await new Promise(() => setInterval(() => {}, 1000));
```
pepoviola, line 102: I think we don't need a promise here.
```suggestion
setInterval(() => {}, 1000);
```

javascript/packages/orchestrator/src/orchestrator.ts
```diff
@@ -161,6 +162,9 @@ export async function start(
     if (networkSpec.settings.node_spawn_timeout)
       client.timeout = networkSpec.settings.node_spawn_timeout;
     network = new Network(client, namespace, tmpDir.path);
+    if (options?.setGlobalNetwork) {
+      options.setGlobalNetwork(network);
+    }
```
pepoviola, line 167: I think you can just do
```suggestion
    options?.setGlobalNetwork(network);
```

javascript/packages/orchestrator/src/providers/podman/podmanClient.ts
```diff
@@ -297,10 +300,13 @@ export class PodmanClient extends Client {
...
-        `\n ${decorators.red("Error: ")} \t ${decorators.bright(error)}\n`,
-      );
-      throw error;
+      if (!this.isTearingDown) {
+        console.log(
+          `\n ${decorators.red("Error: ")} \t ${decorators.bright(error)}\n`,
+        );
+        throw error;
+      }
+      return { exitCode: 0, stdout: "" };
```
pepoviola, line 309: We can use the same pattern that we already use in the `native` provider, and set `allowFail` in the caller.

```suggestion
      debug(error);
      if (!opts?.allowFail) throw error;

      const { exitCode, stdout, message: errorMsg } = error;

      return {
        exitCode,
        stdout,
        errorMsg,
      };
```

javascript/packages/orchestrator/src/orchestrator.ts
l0r1s: Just tried and Typescript gives me:

```Cannot invoke an object which is possibly 'undefined'.```

l0r1s: Got ```Property 'b' does not exist on type 'never'.``` if I try to copy the code in the project to use our TS config

javascript/packages/orchestrator/src/providers/podman/podmanClient.ts
l0r1s: This part is for handling the errors possibly thrown by previous call of runCommand, not the actual calls in the destroyNamespace. In that case, it keeps the behavior of throwing if we are not tearing down the network.

pepoviola (approved):

pepoviola:
thanks!!

PR #1026, opened 2023-05-09
added observability

emamihe:
This piece of code expose prometheus metrics to have visibility on how many zombienet namespaces are running, how many pods are running in each. and what are the namespaces that are older than 2 days that needs to be investigated why they haven't removed.
It's also worth to mention that the alerting has set for notifying us about the namespaces older than 2 days


monitoring/app.go
```diff
@@ -0,0 +1,92 @@
...
+
+			if ns.CreationTimestamp.Time.Before(twoDaysAgo) {
+				zombieNamespaceOld.WithLabelValues(ns.Name).Set(1)
+			} else {
+				zombieNamespaceOld.WithLabelValues(ns.Name).Set(0)
+			}
+		}
+	}
+
+	zombieNamespaceCount.Set(float64(zombieNamespaceCounter))
```
pepoviola, line 91: why we are using `float` for counting?

pepoviola (approved):

PR #1069, opened 2023-05-19
nix maintainance

dzmitry-lahoda:
- binary release depends on some specific cpp libs in rpath, and patching rpath with what I have segmentations fauls or pkg not found from zn insides, so hard to find exact version of cpp stuff which works
- it was needed to run on alpine and nixos, not fixed
- hence fixed previous approach which was run
- disabled husky in nix (it tries download in isolated build)  https://www.youtube.com/watch?v=RAelLqnnOp0
- moved logs to tmp home created by nix (or it fails trying to write immutable to dependency cache)
- linked agains well know cpp libs
- being explicit about copying files in dist (by default as per docs it is all * files, so reduced set of dist files)


pepoviola (approved):

wirednkod:
@dzmitry-lahoda what is the status of this one? Is it ready for review?

dzmitry-lahoda:
i guess i will finalize update as soon as we will migrate to polkdot 1.0.0 deps and friends, with assets hub renaming support

dzmitry-lahoda:
cc @pepoviola

PR #1083, opened 2023-05-22
Fix(orchestrator): `minus` assertion

pepoviola:
Refactor the `getCalcMetric` for use internally `getMetric` and fix the `timeout` issue.
Thanks!

### DIFF
--- a/javascript/packages/orchestrator/src/networkNode.ts
+++ b/javascript/packages/orchestrator/src/networkNode.ts
@@ -293,36 +293,35 @@ export class NetworkNode implements NetworkNodeInterface {
   }
 
   async getCalcMetric(
-    rawMetricName_a: string,
-    rawMetricName_b: string,
-    math_op: string,
+    rawMetricNameA: string,
+    rawMetricNameB: string,
+    mathOp: string,
     comparator: string,
     desiredMetricValue: number,
     timeout = DEFAULT_INDIVIDUAL_TEST_TIMEOUT,
-  ): Promise<number | undefined> {
+  ): Promise<number> {
     let value;
-    let timedout = false;
+    let timedOut = false;
     try {
       const mathFn = (a: number, b: number): number => {
-        return math_op === "Minus" ? a - b : a + b;
+        return mathOp === "Minus" ? a - b : a + b;
       };
 
       const getValue = async () => {
-        let done = false;
-        while (!done && !timedout) {
+        while (!timedOut) {
           const [value_a, value_b] = await Promise.all([
-            this.getMetric(rawMetricName_a),
-            this.getMetric(rawMetricName_b),
+            this.getMetric(rawMetricNameA),
+            this.getMetric(rawMetricNameB),
           ]);
           value = mathFn(value_a as number, value_b as number);
           if (
             value !== undefined &&
             compare(comparator, value, desiredMetricValue)
           ) {
-            done = true;
+            break;
           } else {
             debug(
-              `current values for: [${rawMetricName_a}, ${rawMetricName_b}] are [${value_a}, ${value_b}], keep trying...`,
+              `current values for: [${rawMetricNameA}, ${rawMetricNameB}] are [${value_a}, ${value_b}], keep trying...`,
             );
             await new Promise((resolve) => setTimeout(resolve, 1000));
           }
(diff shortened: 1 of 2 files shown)


javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -209,7 +209,7 @@ export class NetworkNode implements NetworkNodeInterface {
 
   async getMetric(
     rawMetricName: string,
-    comparator: string,
+    comparator?: string,
```
l0r1s, line 212:
```suggestion
    comparator: string,
```

to avoid the non-null assertion

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -292,6 +292,73 @@ export class NetworkNode implements NetworkNodeInterface {
     }
   }
 
+  async getCalcMetric(
+    rawMetricName_a: string,
+    rawMetricName_b: string,
+    math_op: string,
```
l0r1s, line 298: Mix between camelCase and snake_case, could be operandLeft and operandRight

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -292,6 +292,73 @@ export class NetworkNode implements NetworkNodeInterface {
...
+  ): Promise<number | undefined> {
+    let value;
+    let timedout = false;
+    try {
+      const mathFn = (a: number, b: number): number => {
+        return math_op === "Minus" ? a - b : a + b;
+      };
+
+      const getValue = async () => {
+        let done = false;
```
l0r1s, line 311: Not needed given you don't do any operation after the done

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -292,6 +292,73 @@ export class NetworkNode implements NetworkNodeInterface {
...
+          const [value_a, value_b] = await Promise.all([
+            this.getMetric(rawMetricName_a),
+            this.getMetric(rawMetricName_b),
+          ]);
+          value = mathFn(value_a as number, value_b as number);
+          if (
+            value !== undefined &&
+            compare(comparator, value, desiredMetricValue)
+          ) {
+            done = true;
```
l0r1s, line 322:
```suggestion
            break;
```

javascript/packages/orchestrator/src/test-runner/assertions.ts
```diff
@@ -89,34 +89,22 @@ const CalcMetrics = ({
...
-        metric_name_b!,
-        toChaiComparator(op!),
-        target_value!,
-        timeout || DEFAULT_INDIVIDUAL_TEST_TIMEOUT,
+    const results = await Promise.all(
+      nodes.map((node: any) =>
+        node.getCalcMetric(
+          metric_name_a,
+          metric_name_b,
+          math_ops,
```
l0r1s, line 100: I think we should avoid mixing camelCase and snake_case in the code and use a standard

javascript/packages/orchestrator/src/test-runner/assertions.ts
```diff
@@ -89,34 +89,22 @@ const CalcMetrics = ({
...
+        ),
       ),
     );
-
-    const values = await Promise.all([...promise_a, ...promise_b]);
-
-    for (let i = 0; i < nodes.length; i++) {
-      const value = mathFn(values[i], values[i + nodes.length]);
+    for (const value of results) {
       comparatorFn(value as number, target_value as number);
```
l0r1s, line 108: If we need to cast, it's probably mean the type returned is the right one from the functions

javascript/packages/orchestrator/src/networkNode.ts
pepoviola: This allow us to just get the metric without `comparing` since in this case we want just get the value to operate.
Thanks

javascript/packages/orchestrator/src/networkNode.ts
```diff
@@ -231,7 +231,7 @@ export class NetworkNode implements NetworkNodeInterface {
       if (value !== undefined) {
         if (
           desiredMetricValue === null ||
-          compare(comparator, value, desiredMetricValue)
+          compare(comparator!, value, desiredMetricValue)
```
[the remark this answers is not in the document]
pepoviola: To allow get the metric without compare.

wirednkod (approved):

PR #1119, opened 2023-06-12
Fix error when staking bond amount in chainspec is a bigint

nathanwhit:
Currently if you have a chainspec with a staking bond amount that is large enough to be serialized as a `bigint`, you'll
hit an error in zombienet when calculating the balance for validator nodes:

```
Fail to add balance for nodes: [object Object],[object Object],[object Object],[object Object]

 Unexpected error:       TypeError: Cannot mix BigInt and other types, use explicit conversions
```

This is because the calculation is performed assuming the `stakingBond` is a `number` when it can, in fact, be a `bigint` in the chain spec.

This PR fixes this by performing the balance calculation entirely using `bigint`s.

### DIFF
--- a/javascript/packages/orchestrator/src/chainSpec.ts
+++ b/javascript/packages/orchestrator/src/chainSpec.ts
@@ -16,7 +16,7 @@ const debug = require("debug")("zombie::chain-spec");
 const JSONStream = require("JSONStream");
 
 // track 1st staking as default;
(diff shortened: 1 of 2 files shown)


wirednkod (approved):

l0r1s (approved):

pepoviola:
thanks @nathanwhit!

PR #1171, opened 2023-07-04
Fix race conditions with getRandomPort by storing previous ports

tmpolaczyk:
I noticed that sometimes when starting a network with many parachains, one of the nodes seems to fail to start, with the following error seen in the zombienet logs repeatedly:

```
Error 	 fetching metrics from: http://127.0.0.1:44539/metrics
```

After inspecting the node logs, it turns out that the prometheus port of one node was being used as the rpc port of another node. Surprisingly, that random port was assigned by zombienet, as you can see in the yaml files:

```
Collator11.yaml
38:      --prometheus-port 44539 --rpc-port 35281 --ws-port 34043 --collator --

Collator02.yaml
38:      --prometheus-port 35579 --rpc-port 44539 --ws-port 45055 --collator --
```

So that error can be avoided by storing the ports returned by the `getRandomPort` function, and ensuring that it doesn't return the same port twice. So that's what is done in this PR, and I have verified it fixes the problem.

### DIFF
--- a/javascript/packages/utils/src/net.ts
+++ b/javascript/packages/utils/src/net.ts
@@ -7,7 +7,24 @@ import { finished } from "stream/promises";
 import { ReadableStream } from "stream/web";
 import { decorators } from "./colors";
 
-export async function getRandomPort(): Promise<number> {
+const usedPorts = new Map<number, number>();
+
+export interface GetRandomPortOptions {
+  maxRetries?: number;
+  timeout?: number;
+}
+
+/**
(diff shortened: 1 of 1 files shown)


wirednkod:
Thank you @tmpolaczyk . Good catch!

tmpolaczyk:
Fixed the eslint errors.

pepoviola (approved):
thanks!!

pepoviola:
Merging, thanks!

PR #1207, opened 2023-07-29
remove env var from image, to allow other to use easily

pepoviola:
Remove `env var` from image, we should set it in our infra and not in the public image.


wirednkod:
While I agree, this seems to break things - should we first make sure that our infra is updated correctly prior to merging this one?

pepoviola:
> While I agree, this seems to break things - should we first make sure that our infra is updated correctly prior to merging this one?

The change should be at repo level ( in the pipeline's files), so this will no break anything but we need a follow up pr in polkadot/cumulus/substrate.

Thx!

wirednkod (approved):

PR #1235, opened 2023-08-09
Add configurable key types

### LINKED ISSUE #1185: Key type insertion selection
As of today, zombienet injects a bunch of key types in the key store to stay compatible with most of the chains, e.g.:

- `nimbus`
- `aura`
- `vrf`
- `granda`

It would be good if we could configure which of those keys we want to be inserted in the keystore. For instance we could have a flag in the 

`accepted_key_types: ["nimbus", "vrf"]
`
which would only insert `nimbus` and `vrf` keys in the keystore.

wirednkod:
Pending TODOs: 

- [x] Add tests 
- [ ] Add docs 

Fixes #1185

### DIFF
--- a/examples/check_creations.sh
+++ b/examples/check_creations.sh
@@ -1,0 +1,28 @@
+#!/bin/bash
+path=`pwd`
+some=${path##*/}
+cd keystore
+
+if [ $some = "alice" ];
+then
+    ## alice has all key types (11) thus we just count the number of files in keystore dir
+    count=$(ls | wc -l)
+    if [ $count = 11 ]; then
+        exit 0
+    fi
+elif [ $some = "bob" ];
+then
+    result=`ls $(echo -n aura | hexdump -v -e '/1 "%02X"')*`
+    count=$(echo -e "$result" | wc -l)
+    if [ $count = 1 ]; then
+        exit 0
+    fi
+elif [ $some = "collator01" ];
+then
+    result=`ls $(echo -n gran | hexdump -v -e '/1 "%02X"')*`
+    count=$(echo -e "$result" | wc -l)
+    exit 0
+fi
+
+exit 1
+echo "ERRORED"
--- a/examples/check_creations.sh
+++ b/examples/check_creations.sh
@@ -24,5 +24,5 @@ then
     exit 0
 fi
 
-exit 1
 echo "ERRORED"
+exit 1
--- a/examples/check_creations.sh
+++ b/examples/check_creations.sh
@@ -1,27 +1,29 @@
 #!/bin/bash
-path=`pwd`
-some=${path##*/}
-cd keystore
 
-if [ $some = "alice" ];
+KEY_ENCODED=$1
+SCRIPT=$(realpath "$0")
+SCRIPTPATH=$(dirname "$SCRIPT")
+
+if [[ $SCRIPTPATH == "/tmp" ]]; then
+    cd ../data
+else
+    # native provider use `root dir` to run the script
+    cd data
+fi
+
+if [ $KEY_ENCODED = "ALL" ];
 then
-    ## alice has all key types (11) thus we just count the number of files in keystore dir
+    ## should have all key types (11) thus we just count the number of files in keystore dir
+    cd $(find . -name keystore)
     count=$(ls | wc -l)
-    if [ $count = 11 ]; then
+    if [ $count -eq 11 ]; then
         exit 0
     fi
-elif [ $some = "bob" ];
-then
-    result=`ls $(echo -n aura | hexdump -v -e '/1 "%02X"')*`
-    count=$(echo -e "$result" | wc -l)
-    if [ $count = 1 ]; then
+else
+    count=$(find . -name "${KEY_ENCODED}*" |wc -l)
+    if [ $count -eq 1 ]; then
         exit 0
     fi
-elif [ $some = "collator01" ];
-then
-    result=`ls $(echo -n gran | hexdump -v -e '/1 "%02X"')*`
-    count=$(echo -e "$result" | wc -l)
-    exit 0
 fi
 
 echo "ERRORED"
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -42,6 +42,20 @@ import {
   envVars,
 } from "./types";
 
+const default_keystore_key_types: string[] = [
+  "aura",
+  "babe",
+  "imon",
+  "gran",
+  "audi",
+  "asgn",
+  "para",
+  "beef",
+  "nmbs",
+  "rand",
+  "rate",
+];
+
 const debug = require("debug")("zombie::config-manager");
 
 // get the path of the zombie wrapper
@@ -119,6 +133,9 @@ export async function generateNetworkSpec(
       defaultImage: config.relaychain.default_image || DEFAULT_IMAGE,
       defaultCommand: config.relaychain.default_command || DEFAULT_COMMAND,
       defaultArgs: config.relaychain.default_args || [],
+      default_keystore_key_types:
+        config.relaychain.default_keystore_key_types ||
+        default_keystore_key_types,
       randomNominatorsCount: config.relaychain?.random_nominators_count || 0,
       maxNominations:
         config.relaychain?.max_nominations || DEFAULT_MAX_NOMINATIONS,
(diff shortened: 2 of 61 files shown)


javascript/packages/orchestrator/src/configGenerator.ts
```diff
@@ -42,6 +42,20 @@ import {
   envVars,
 } from "./types";
 
+const default_keystore_key_types: string[] = [
```
pepoviola, line 45: We can move this list to `const` an reuse here and https://github.com/paritytech/zombienet/pull/1235/files#diff-b1dd06ee7d2b94fe91604cba031f78c09a937a5f71cd659fa4baf7601ac9ecfbR94, so we maintain only one list of possible keys.

wirednkod:
For parachains we are adding the [parachain.name](https://github.com/paritytech/zombienet/blob/7fd7b94391485336a1610ad1bbb9795f75c65173/javascript/packages/orchestrator/src/spawner.ts#L85) (which is the id of the parachain) for storing the key_types which creates a completely different directory than the one where the rest of the directories/files are saved (data, cfg etc etc);

As a result if that line is NOT commented out all files for the collator01 are saved under the path:

> /var/folders/7q/3ws1fn9d0pn961zh9fk5vgm40000gn/T/zombie-735a4dbb8bf681322ab93e0329b6f732_-42616-8LMRjq8GPabq/collator01

but the keystores are saved under the path (has an extra `/100`) :

> /var/folders/7q/3ws1fn9d0pn961zh9fk5vgm40000gn/T/zombie-735a4dbb8bf681322ab93e0329b6f732_-42616-8LMRjq8GPabq/100/collator01

This feels like a bug to me, and I have commented out that line;
Please let me know if that decision is correct or I should revert that change and try to figure out why that extra parachain name is added;

examples/check_creations.sh
```diff
@@ -0,0 +1,28 @@
...
+    fi
+elif [ $some = "collator01" ];
+then
+    result=`ls $(echo -n gran | hexdump -v -e '/1 "%02X"')*`
+    count=$(echo -e "$result" | wc -l)
+    exit 0
+fi
+
+exit 1
+echo "ERRORED"
```
pepoviola, line 28: this line will never be printed since we are exit before.

pepoviola (commented):
looks good!! just an mall `nit`

wirednkod:
@pepoviola what about the line of code that i propose to be removed?

javascript/packages/orchestrator/src/keys.ts
```diff
@@ -85,6 +92,23 @@ export async function generateKeystoreFiles(
...
+  node.keystore_key_types?.forEach((key_type: string) => {
+    console.log(
+      "!!!! node.keystore_key_types",
+      key_type,
+      node.keystore_key_types,
+    );
+    console.log(
+      "!!!! node.keystore_key_types",
+      DEFAULT_KEYSTORE_KEY_TYPES.includes(key_type),
+    );
```
pepoviola, line 104:
```suggestion
```
remove `console.log`

pepoviola:
> @pepoviola what about the line of code that i propose to be removed?

Ahh, let me re-check. Thanks!

javascript/packages/orchestrator/src/spawner.ts
```diff
@@ -82,7 +82,7 @@ export const spawnNode = async (
   if (node.accounts) {
     // check if the node directory exists if not create (e.g for k8s provider)
     let nodeFilesPath = client.tmpDir;
-    if (parachain && parachain.name) nodeFilesPath += `/${parachain.name}`;
+    // if (parachain && parachain.name) nodeFilesPath += `/${parachain.name}`;
```
pepoviola, line 85: Sorry I miss this change, the idea behind using the the `parachains name` is to group all the artifacts of the parachain in the same `directory`. Also, worth nothing that this is the `local` path used in the spawn phase and not the real `keystore` in the node.

examples/check_creations.sh
pepoviola: Also, I think this test will only work using the `native` provider. Since the path of the keystore inside the `pods` is different.

javascript/packages/orchestrator/src/keys.ts
```diff
@@ -69,7 +70,13 @@ export async function generateKeystoreFiles(
   await makeDir(keystoreDir);
 
   const paths: string[] = [];
-  const keysHash = {
+
+  interface DefaultKeystoreKeyTypes {
+    [key: string]: string;
+  }
+  let keysHash: DefaultKeystoreKeyTypes = {};
```
l0r1s, line 77:
```suggestion
  let keysHashes: DefaultKeystoreKeyTypes = {};
```

javascript/packages/orchestrator/src/keys.ts
```diff
@@ -85,6 +92,23 @@ export async function generateKeystoreFiles(
...
+      "!!!! node.keystore_key_types",
+      key_type,
+      node.keystore_key_types,
+    );
+    console.log(
+      "!!!! node.keystore_key_types",
+      DEFAULT_KEYSTORE_KEY_TYPES.includes(key_type),
+    );
+    if (DEFAULT_KEYSTORE_KEY_TYPES.includes(key_type))
+      keysHash[key_type] = default_keystore_key_types[key_type];
```
l0r1s, line 106:
```suggestion
      keysHashes[key_type] = default_keystore_key_types[key_type];
```

javascript/packages/orchestrator/src/keys.ts
```diff
@@ -85,6 +92,23 @@ export async function generateKeystoreFiles(
...
+  });
+
+  if (!keysHash || Object.keys(keysHash).length === 0)
```
l0r1s, line 109: ```keysHashes``` can't be ```null```/```undefined``` because it's defined above and it's a empty object.

```suggestion
  if (Object.keys(keysHashes).length === 0)
```

l0r1s (commented):
Lgtm, some only one dead condition + renaming

javascript/packages/orchestrator/src/spawner.ts
wirednkod: Yes. My point here is that this path `nodeFilesPath += `/${parachain.name}'` is only created for the keystores which makes us end up with:
```
----> zombie-random-dir/collator01/(all dirs and configs etc)
          zombie-random-dir/100/collator01/keystores/<the keystores only>
```
which to me makes not much of sense - as just the keystores are separated.

In addition since this is just the local path - I do not see any added value on having it separated (since we are the only ones using it - and not the node

examples/check_creations.sh
wirednkod: Can you elaborate a this bit more? All the paths I use inside are relevant (initiated by `pwd`). I will continue testing on podman and k8s but I think if we have these inconsistencies between providers maybe we should create a "path mapping" between supported providers in order to be able to be consistent

l0r1s: I think what @pepoviola means is that the ```exit 1``` is like a "return" so the rest of the script after it (the ```echo "ERRORED"``` in that case) won't be executed

javascript/packages/orchestrator/src/spawner.ts
pepoviola: Hey @wirednkod, the main reason of this separation is that in the `native` provider we create a new directory for each node as isolated environment that `acts` as node root directory. So, you can think of `zombie-random-dir/collator01/(all dirs and configs etc)` as the `isolated running node`, and the other one is the directory we use to store all the created artifacts needed to run the `node`. So, in native they can look similar (and have dups files) but that is because from zombienet perspective they are different.

Make sense? Thanks!!

pepoviola, line 85:
```suggestion
    if (parachain && parachain.name) nodeFilesPath += `/${parachain.name}`;
```

examples/check_creations.sh
pepoviola: Yes, we follow the same directories structure in all the providers in the `nodes`

```
/data --> we use as base-path for running the node, here you will find the db, keystore, etc.
/relay-data --> Same as `/data` but only used in collators as base-path for the full-node of the relaychain.
/cfg --> we store the config and extra scripts, here you will find the chain-spec, and all the scripts we upload as part of the test-runner.
```

In the case of `native` provider the `tmpDir` acts as the `/` of the `node` fs.

pepoviola (approved):
thanks!!

PR #1236, opened 2023-08-11
Fix 404 link of moonbeam

wirednkod:


javascript/yarn.lock
l0r1s, line 1: Why is there a yarn.lock added ? I think it must be a mistake

wirednkod: yes! this is a mistake.. probably erroneously run `yarn` instead of `npm`

l0r1s (approved):

PR #1247, opened 2023-08-15
fix(orchestrator): set undying/adder collators not validators by default

### LINKED ISSUE #1245: Fix collator (undying/adder) cmd args
Followup from https://github.com/paritytech/polkadot/pull/7617#issuecomment-1678536346, we should not use `--validator` in those collators cmd.

pepoviola:
fix #1245. Don't set `undying/adder` collators as validators by default.

### DIFF
--- a/javascript/packages/orchestrator/src/cmdGenerator.ts
+++ b/javascript/packages/orchestrator/src/cmdGenerator.ts
@@ -270,7 +270,7 @@ export async function genCmd(
     const parachainIdArgIndex = args.findIndex((arg) =>
       arg.includes("--parachain-id"),
     );
-    if( parachainIdArgIndex >=0) args.splice(parachainIdArgIndex, 1);
+    if (parachainIdArgIndex >= 0) args.splice(parachainIdArgIndex, 1);
     args.push(`--parachain-id ${parachainId}`);
   }
 
--- a/javascript/packages/orchestrator/src/cmdGenerator.ts
+++ b/javascript/packages/orchestrator/src/cmdGenerator.ts
@@ -270,7 +270,7 @@ export async function genCmd(
     const parachainIdArgIndex = args.findIndex((arg) =>
       arg.includes("--parachain-id"),
     );
-    args.splice(parachainIdArgIndex, 1);
+    if( parachainIdArgIndex >=0) args.splice(parachainIdArgIndex, 1);
     args.push(`--parachain-id ${parachainId}`);
   }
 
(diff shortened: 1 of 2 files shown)


javascript/packages/orchestrator/src/configGenerator.ts
```diff
@@ -529,11 +529,20 @@ async function getCollatorNodeFromConfig(
...
   const externalPorts = await getExternalPorts(provider, ports, collatorConfig);
 
+  // By default cumulus collators are `validators`, this implies that we will
+  // run those with this flag `--collator`.
+  const isValidator =
+    collatorConfig.validator !== undefined
+      ? collatorConfig.validator
+      : cumulusBased
+      ? true
+      : false;
```
mrcnski, line 539: I'm having a hard time parsing this nested ternary statement. Also, I think JS has a `??` operator or something similar you could use, IIRC.

javascript/packages/orchestrator/src/cmdGenerator.ts
```diff
@@ -270,7 +270,7 @@ export async function genCmd(
     const parachainIdArgIndex = args.findIndex((arg) =>
       arg.includes("--parachain-id"),
     );
-    args.splice(parachainIdArgIndex, 1);
+    if( parachainIdArgIndex >=0) args.splice(parachainIdArgIndex, 1);
```
wirednkod, line 273:
```suggestion
    if( parachainIdArgIndex >=0 ) args.splice(parachainIdArgIndex, 1);
```

javascript/packages/orchestrator/src/configGenerator.ts
wirednkod: Correct. I think this could simplier be:
```suggestion
  const isValidator = !!collatorConfig.validator || !!cumulusBased;
```

pepoviola: Sounds good to me, old ternary addiction 😂

we can also use 
```
const isValidator = Boolean(collatorConfig.validator) || Boolean(cumulusBased);
```

Thanks!!

javascript/packages/orchestrator/src/cmdGenerator.ts
pepoviola: Is this something that the `linting` should `fix` or at least catch?

pepoviola: fixed by linter.

javascript/packages/orchestrator/src/configGenerator.ts
pepoviola: I just add a better comment about this ternary, resolving here.
Thx!

wirednkod (approved):

PR #1252, opened 2023-08-16
Update reviewers of zombienet packages (automate reviewers)

wirednkod:


.github/CODEOWNERS
```diff
@@ -2,3 +2,6 @@
 /.github/ @paritytech/ci @pepoviola
 /.gitlab-ci.yml @paritytech/ci @pepoviola
 /scripts/ci/ @paritytech/ci @pepoviola
```
pepoviola, line 4: Can we add @emamihe here? Thx!

wirednkod: Ah sure. I was not sure if we should annoy him automatically for non infra issues

pepoviola (approved):

PR #1298, opened 2023-08-31
Fix exiting of processes;

### LINKED ISSUE #1291: [bug] invalid chain-spec path should exit as error
Check  https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/3492048
If the path provided is invalid we should exit with an error code.

wirednkod:
Fixes #1291

### DIFF
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -169,7 +169,7 @@ export async function generateNetworkSpec(
           `Genesis spec provided does not exist: ${chainSpecPath}`,
         ),
       );
-      process.exit();
+      process.exit(1);
     } else {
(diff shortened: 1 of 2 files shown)


javascript/packages/orchestrator/src/configGenerator.ts
```diff
@@ -337,7 +337,7 @@ export async function generateNetworkSpec(
               `Genesis spec provided does not exist: ${genesisStatePath}`,
```
pepoviola, line 337:
```suggestion
              `Genesis state file path provided does not exist: ${genesisStatePath}`,
```

pepoviola: typo not related to the fix :)

pepoviola (approved):

PR #1316, opened 2023-09-04
chore(nix): maintenance and cleanup

dzmitry-lahoda:
- binary release depends on some specific cpp libs in rpath, and patching rpath with what I have segmentations fauls or pkg not found from zn insides, so hard to find exact version of cpp stuff which works
- it was needed to run on alpine and nixos, not fixed
- hence fixed previous approach which was run
- disabled husky in nix (it tries download in isolated build)  https://www.youtube.com/watch?v=RAelLqnnOp0
- moved logs to tmp home created by nix (or it fails trying to write immutable to dependency cache)
- linked agains well know cpp libs
- being explicit about copying files in dist (by default as per docs it is all * files, seems nix builder does not respect defaults - there is bug)


dzmitry-lahoda:
cc @pepoviola

dzmitry-lahoda:
I run ` npm run lint:write` bit it fixed unrelated files to my commit. So doing nothing.

wirednkod (approved):
Thnk you @dzmitry-lahoda

wirednkod:
> I run ` npm run lint:write` bit it fixed unrelated files to my commit. So doing nothing.

It is stupid (I know) but can you add 1 empty line at the end of the main `package.json` please?

javascript/package.json
```diff
@@ -54,4 +57,4 @@
     "packages/orchestrator",
     "packages/cli"
   ]
-}
+}
```
wirednkod, line 60:
```suggestion
}

```

pepoviola:
Thanks @dzmitry-lahoda! Couple of questions:
- I saw that we are creating an `example` config and download `polkadot/polkadot-parachain` as part of this changes, this will only works on `linux_x86-64` since the release binaries are built for that os/arch. Do we manage to not download extra binaries in other OS/arch?
- Should be a way to easily update hashes from newest releases and automate the process as part of our pipeline (maybe some guide here will be appreciated)
- We should include the `uninstall` steps as part of the docs.

Thanks again for help us with this!

dzmitry-lahoda:
> this will only works on linux_x86-64 since the release binaries are built for that os/arch

i will work with official parity releases, as long as parity does not release binaries for arm/mac it will not work.

dzmitry-lahoda:
> Do we manage to not download extra binaries in other OS/arch?

this is possible by making conditional package, sure

dzmitry-lahoda:
making example conditional, as it is clear how to do it.

pepoviola (approved):
thanks @dzmitry-lahoda 🚀

pepoviola:
> > We should include the uninstall steps as part of the docs.
> 
> not sure I get it. i added nothing new about installation. nothing from nix is installed, not shell, not example, nothing. I mean if somebody wants to install sure. but runs do not install.
> 
> UPDATE: i am not sure how it relates to this PR tbh

Oks, yes my question wasn't related to the pr. Thx!

pepoviola:
> image
> 
> imho process-compose is awesome, works on win/mac/*nix. if ever ZN will use it instead of own manual orchestrate would be awesome (PC has REST and CLI api - so super easy to interact). and in theory it allows to use partial isolation (isolation without building full docker image). just FYI
> 
> on *nix PC shows all opened ports.
> 
> I am DONE

Looks great, I will explore it a little bit more with the team. Thanks!

pepoviola:
Merging, thanks @dzmitry-lahoda

PR #1466, opened 2023-11-01
Adding gitspiegel-trigger workflow

mutantcornholio:
Using a workflow to trigger mirroring instead of a webhook allows us to reuse "Approving workflow runs from public forks" GitHub feature to somewhat protect us from malicious PRs

### DIFF
--- a/.github/workflows/gitspiegel-trigger.yml
+++ b/.github/workflows/gitspiegel-trigger.yml
@@ -1,0 +1,20 @@
+name: gitspiegel sync
+
+on:
+  pull_request:
+    types:
+      - opened
+      - synchronize
+      - unlocked
+      - ready_for_review
+      - reopened
+
+jobs:
+  sync:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Trigger sync via API
+        run: |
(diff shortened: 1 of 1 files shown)


wirednkod (changes requested):
Thank you @mutantcornholio 
Can you please elaborate a bit more on the PR, e.g.  what this change do, how is it used - where this secret is set etc etc?

mutantcornholio:
GitHub has a setting that requires manual click for executing GHA on the branch, for the first-time contributors: https://docs.github.com/en/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks.

After this PR, gitspiegel will respect that setting.
So, for PRs from first-time contributors, gitspiegel won't do mirroring until the button in PR is clicked.   
More info: https://github.com/paritytech/gitspiegel/issues/169

The secret is being set per repo, currently by security.

wirednkod (approved):
Thank you @mutantcornholio 

cc @pepoviola can you have a look as well please?

pepoviola (approved):

pepoviola:
thanks @mutantcornholio!!

Issue #1449, opened 2023-10-24
Add support for `polkadot-dev`


SBalaguer:
### Is your feature request related to a problem? Please describe.

In the past when the runtime was part of the release, one could run a local version of Polkadot buy passing the flag `--chain polkadot-dev`. Now, however, the Polkadot Runtime is being released from the fellowship repositorty (https://github.com/polkadot-fellows/runtimes), and therefore the Substrate Node and the Polkadot Runtime are released from different places.

### Describe the solution you'd like

Ideally I could run a local instance of Polkadot with Zombienet for testing purposes.

pepoviola:
related to https://github.com/paritytech/polkadot-sdk/pull/1256, ping @michalkucharczyk there is a way to generate the spec currently in master? I saw that the `pr` add a new util to generate but I'm not sure if there is another way from the current master branch.

michalkucharczyk:
Bringing back chain-specs for kusama and polkadot is planned in: https://github.com/paritytech/polkadot-sdk/issues/1984
As discussed, it can now partially be done with work provided in https://github.com/paritytech/polkadot-sdk/pull/1256:
- it is possible to get default config for the wasm blob,
- it is possible to verify and build the raw chain spec from the plain chain spec,

Some internal structures of the chain spec (e.g. the layout of seession keys) needs to be temporarily hardcoded somewhere.

SBalaguer:
Thank you both for the answers!

I'm tagging @carlosala who works at @zondax and needs this functionality to be able to test the new ledger app with the changes the new runtime will have.

With [Referenda #297](https://kusama.polkassembly.io/referenda/297) being voted on Kusama, this is a bit time critical now.

For the moment, Carlo needs to be able to actually test the same runtime as Kusama locally with Zombienet. Two validator nodes would be more than enough in this case. Thanks! :))

michalkucharczyk:
there is also a tool, added by @bkchr , this may help:
https://github.com/polkadot-fellows/runtimes/pull/78

pepoviola:
Hi @SBalaguer / @carlosala, as @michalkucharczyk shared you can use that branch to use the `chain-spec-generator` tool.
The steps to run a 2 validators (alice/bob) chain using kusama-local with zombienet are:

1 - Generate the raw chain spec, from this [branch](https://github.com/polkadot-fellows/runtimes/tree/bkchr-chain-spec-generator) of the runtime repo with 
```
[...]

PR #1500, opened 2023-11-09
Use chain_spec_path to detect asset hub polkadot

svyatonik:
There's no way now to use `polkadot-parachain` with `--chain=asset-hub-polkadot-local`, so we need to use chain spec path. But zombienet only uses the `chain` parameter to detect polkadot AH and generate proper keystore files (for which workaround is to explicitly specify `aura_ed` in `keystore_key_types`) + properly modify chain validators in spec (for which there's no easy workaround).

### DIFF
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -242,7 +242,7 @@ export async function generateNetworkSpec(
 
   if (config.parachains && config.parachains.length) {
     for (const parachain of config.parachains) {
-      const para: CHAIN = whichChain(parachain.chain || "");
+      const para: CHAIN = whichChain(parachain.chain || parachain.chain_spec_path || "");
 
       let computedStatePath,
         computedStateCommand,
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -242,7 +242,9 @@ export async function generateNetworkSpec(
 
   if (config.parachains && config.parachains.length) {
     for (const parachain of config.parachains) {
-      const para: CHAIN = whichChain(parachain.chain || parachain.chain_spec_path || "");
+      const para: CHAIN = whichChain(
+        parachain.chain || parachain.chain_spec_path || "",
+      );
 
       let computedStatePath,
         computedStateCommand,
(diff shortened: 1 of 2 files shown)


pepoviola:
Hi @svyatonik thanks for your contribution, I think this is related to the new `runtimes` repo right? We add support for using a `custom external tool`  to build the chain-spec [here](https://github.com/paritytech/zombienet/pull/1471).
If this is the case you can follow the instructions from https://github.com/paritytech/zombienet/issues/1449#issuecomment-1789054394 and use the `chain-spec-generator` tool from the runtimes repo.

Please let me know if this cover your case?
Thanks!

svyatonik:
Nope, that doesn't solve our problem :) I'm also using that tool, but because there were no support of `chain_spec_generator` for parachains, I had to write a script that emulated old `polkadot-parachain` behavior and either starts `chain-spec-generator` or the actual `polkadot-parachain` depending on args.

The "core" problem is that the Asset Hub Polkadot is a special case and when zombienet is generating/patching chain spec for this chain, it does two things:
1) it writes [ed keys for Aura](https://github.com/paritytech/zombienet/blob/main/javascript/packages/orchestrator/src/keys.ts#L79) to the keystore;
2) and yet it writes address, associated with [ed key to the chain spec](https://github.com/paritytech/zombienet/blob/main/javascript/packages/orchestrator/src/chain-decorators/asset_hub_polkadot.ts#L8) (validators set).

Without those two points, the local Polkadot Asset Hub simply won't produce blocks (as I said in description, there's a workaround for (1), but no easy workaround for (2)). Before fellowship repo, the `polkadot-parachain` was supporting the `--chain=asset-hub-polkadot-local` argument && everyone used that. But now with AH moved to fellowship, we have to provide external chain spec, so it doesn't work anymore. So the [`parachain.chain` here](https://github.com/paritytech/zombienet/pull/1500/files#diff-3833f9ff37d41f123bcc15a054a60463a70c25e2ed37b9525c948158d4b748bcR246) and [`parachain.chain` here](https://github.com/paritytech/zombienet/pull/1500/files#diff-24b7795ec93992e92b59b879677a7a7e90adcdf07284b6910a165592b680dabfR88) are undefined and this hack doesn't work. So what I suggest is to also check the `chain_spec_path` there, where we could actually use the `asset-hub-polkadot` name to trigger it.

pepoviola (approved):

pepoviola:
Hi @svyatonik, thanks for the reply 👍 . @NachoPal add system parachains in this pr https://github.com/polkadot-fellows/runtimes/pull/81 but isn't yet merged, I will merge this and then we can review if the `chain-spec-generator` works as expected as external generator set in `chain_spec_command`.
Again, thanks for your contribution!!

PR #1439, opened 2023-10-20
Updated required repo

Neyromancer:
## Description

While trying to spawn a network in kubernetes an error related to the use namesapce different then default appeared:

```
 Error:  	 Error: Command failed with exit code 1: kubectl --kubeconfig /Users/wirednkod/.kube/config delete podmonitor zombie-ef4976230e7e09608e35a4bb97d075d0 -n monitoring
Error from server (NotFound): podmonitors.monitoring.coreos.com "zombie-ef4976230e7e09608e35a4bb97d075d0" not found
```

Related [PR #1412](https://github.com/paritytech/zombienet/issues/1412) Error when trying to spawn network in kubernetes

--------------------
## Tasks:

- [ ] Make clear in the docs that you need to be set in the default namespace to start working with zombienet.
- [ ] Print a warning at start that different then default namspace is used

Issue #1412, opened 2023-10-13
Error when trying to spawn network in kubernetes 


wirednkod:
While trying to spawn a network in kubernetes the following error appeared:

```
➜  javascript git:(main) ✗ npm run zombie -- spawn ../tests/chaos/0001-delay.toml

> zombienet@0.1.0 zombie
> node ./packages/cli/dist/cli.js spawn ../tests/chaos/0001-delay.toml

╔════════════════════╤════════════════════════════════════════════════════════════════════════════════════════════════════╗
║ 🧟 Zombienet 🧟    │ Initiation                                                                                         ║
╟────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────╢
║ Provider           │ kubernetes                                                                                         ║
╟────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────╢
║ Namespace          │ zombie-ef4976230e7e09608e35a4bb97d075d0                                                            ║
╟────────────────────┼────────────────────────────────────────────────────────────────────────────────────────────────────╢
║ Temp Dir           │ /var/folders/7q/3ws1fn9d0pn961zh9fk5vgm40000gn/T/zombie-ef4976230e7e09608e35a4bb97d075d0_-40811-d… ║
╚════════════════════╧════════════════════════════════════════════════════════════════════════════════════════════════════╝
Error:  	 Error: Command failed with exit code 1: kubectl --kubeconfig /Users/wirednkod/.kube/config apply -f -
Error from server (NotFound): error when creating "STDIN": namespaces "zombie-b3b08c66fc8e8e48d2876049d4477f7e" not found
Error from server (NotFound): error when creating "STDIN": namespaces "zombie-b3b08c66fc8e8e48d2876049d4477f7e" not found
Error from server (NotFound): error when creating "STDIN": namespaces "zombie-b3b08c66fc8e8e48d2876049d4477f7e" not found
Error from server (NotFound): error when creating "STDIN": namespaces "zombie-b3b08c66fc8e8e48d2876049d4477f7e" not found

undefined
┌────────────────────┬────────────────────────────────────────────────────────────────────────────────────────────────────┐
│ Node'              │ /var/folders/7q/3ws1fn9d0pn961zh9fk5vgm40000gn/T/zombie-ef4976230e7e09608e35a4bb97d075d0_-40811-d… │
│ s logs:            │                                                                                                    │
│ m                  │                                                                                                    │
[...]

PR #1503, opened 2023-11-09
Updated docs/book files with new mdbook version

Neyromancer:
## Descritption

Update docs/book HTML files with new version of mdbook.
mdbook v0.4.35

Relates to [PR #1439](https://github.com/paritytech/zombienet/pull/1439) and [issue #1412](https://github.com/paritytech/zombienet/issues/1412)

This PR should be as a preparation for [PR #1439](https://github.com/paritytech/zombienet/pull/1439) which make changes to doc files.


pepoviola:
Thanks for your contribution @Neyromancer! but let me ping @wirednkod to correct me if I'm wrong, but I think the `html` files of `docs/book` are auto-generated by a workflow in GHA. We should no manually change those, if need to update the mdBook version we should update the [workflow](https://github.com/paritytech/zombienet/blob/main/.github/workflows/gh-pages.yml)
Thx!

Neyromancer:
> Thanks for your contribution @Neyromancer! but let me ping @wirednkod to correct me if I'm wrong, but I think the `html` files of `docs/book` are auto-generated by a workflow in GHA. We should no manually change those, if need to update the mdBook version we should update the [workflow](https://github.com/paritytech/zombienet/blob/main/.github/workflows/gh-pages.yml) Thx!

Looks as if it is indeed should be updated there. Did not check CI scripts, pardon

wirednkod:
> Thanks for your contribution @Neyromancer! but let me ping @wirednkod to correct me if I'm wrong, but I think the `html` files of `docs/book` are auto-generated by a workflow in GHA. We should no manually change those, if need to update the mdBook version we should update the [workflow](https://github.com/paritytech/zombienet/blob/main/.github/workflows/gh-pages.yml?rgh-link-date=2023-11-09T22%3A51%3A04Z) Thx!

That is correct. The changed files will though be overwritten and since same changes applied everywhere then, the final doc files will be correct

wirednkod:
Closing this as it does not make sense to update the files that willbe overridden

pepoviola:
Hi @Neyromancer thanks for your contribution!! We just closed this one since as we said before we should not change `html` files directly since those are generated again from GHA.
Thanks again!

PR #1507, opened 2023-11-10
Add Chaos support to CLI

wirednkod:


crates/parser/src/zombienet.pest
```diff
@@ -27,6 +27,7 @@ span_id = { ASCII_HEX_DIGIT{32} }
 block_height = { "blockheight" | "block height" | "best block" }
 finalized_height = { "finalised height" | "finalised block" }
 peers_count = { "peers count" | "peers" }
+chaos_type = { "latency" | "jitter" | "correlation" }
```
pepoviola, line 30: `correlation` manage how `latency/jitter` works internally, I think most of the cases will be `100` (the default value). To not complicate more the D.S.L I would say that we should only allow to set `latency` and `jitter`.

crates/parser/src/zombienet.pest
```diff
@@ -83,6 +85,10 @@ pause = { node_name ~ "pause" }
 resume = { node_name ~ "resume" }
 restart = { node_name ~ "restart" ~ ("after" ~ int+ ~ seconds)? }
 
+// CHAOS COMMANDS
+setter = { node_name ~ "set" ~ chaos_type ~ "by" ~ int+ ~ (seconds | miliseconds)}
+remover = { node_name ~ "remove" ~ chaos_type }
```
pepoviola, line 90:
```suggestion
chaos_setter = { node_name ~ "set" ~ chaos_type ~ "by" ~ int+ ~ (seconds | miliseconds)}
chaos_remover = { node_name ~ "remove" ~ chaos_type }
```

pepoviola: This will also only allow to set one of the two options (latency/jitter), I think we can work to cover all 3 options in the same command (allow to set latency / jitter / latency and jitter).

PR #1515, opened 2023-11-14
Stop using shell output redirects

JoshOrndorff:
This tiny PR changes the way that genesis wasm and head data are generated very slightly.

```bash
# Old way
collator-node export-genesis-wasm > some-wasm-file

# New way
collator-node export-genesis-wasm some-wasm-file
```

In the old days, using the shell redirect was the only way to get the file you wanted. But later cumulus added the ability to provide an [output file argument](https://github.com/paritytech/polkadot-sdk/blob/master/cumulus/client/cli/src/lib.rs#L223-L225) and just fallback to printing directly to stdout.

I prefer using the argument because sometimes I find myself needing to print debugging output when actually generating that data, and I don't want my debugging output being redirected to the generated file. (As a concrete example, here is a [current PR](https://github.com/paritytech/polkadot-sdk/pull/2331) where I'm having this problem, and it has happened in the past too.)

### DIFF
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -353,7 +353,7 @@ export async function generateNetworkSpec(
           ? parachain.genesis_state_generator
           : `${collatorBinary} ${DEFAULT_GENESIS_GENERATE_SUBCOMMAND}`;
 
-        computedStateCommand += ` > {{CLIENT_REMOTE_DIR}}/${GENESIS_STATE_FILENAME}`;
+        computedStateCommand += ` {{CLIENT_REMOTE_DIR}}/${GENESIS_STATE_FILENAME}`;
       }
 
       if (parachain.genesis_wasm_path) {
@@ -376,7 +376,7 @@ export async function generateNetworkSpec(
           ? parachain.genesis_wasm_generator
           : `${collatorBinary} ${DEFAULT_WASM_GENERATE_SUBCOMMAND}`;
 
-        computedWasmCommand += ` > {{CLIENT_REMOTE_DIR}}/${GENESIS_WASM_FILENAME}`;
+        computedWasmCommand += ` {{CLIENT_REMOTE_DIR}}/${GENESIS_WASM_FILENAME}`;
       }
 
       let parachainSetup: Parachain = {


wirednkod:
Thank you @JoshOrndorff for this. There are more places where the outputs redirect is used and I think a more careful search and discovery of those is needed. I agree with the approach ( cc @pepoviola what do you think), but all of those should be in the same PR IMO.
Some examples:

https://github.com/paritytech/zombienet/blob/main/javascript/packages/orchestrator/src/paras.ts#L239-L242
https://github.com/paritytech/zombienet/blob/main/javascript/packages/orchestrator/src/paras.ts#L259

JoshOrndorff:
Are there build instructions anywhere so that I can build zombienet myself from my branch? It is hard to contribute to it when I can only wait for a release to test my changes.

pepoviola:
> Are there build instructions anywhere so that I can build zombienet myself from my branch? It is hard to contribute to it when I can only wait for a release to test my changes.

Hey @JoshOrndorff, sorry about the delay. I saw in your pr that you are now using [`eprintln`](https://github.com/paritytech/polkadot-sdk/pull/2331/files#diff-2ec59d539a0c6221eee33dcd9de7f7054268afad7afbd58a2fec34f4124c406eR156) and those lines will not be redirected to the file with the current shell redirection `>` .  Is this the case?

But, as you suggest using the new method could be better since others that use `println` and are hit by the same issue.
Related to the `build`, [here](https://paritytech.github.io/zombienet/development.html) are the steps to build from source using `node.js`.

Thanks! and again sorry about the delayed answer.

pepoviola:
Hey @JoshOrndorff, Thanks for open this `pr` but I think we will keep the shell redirect since works across cumulus/adder/undying collators. As I said before if you want to print without goes to the file you can use `eprintln` to print to stderr (as you made in your pr :)).

Again, thanks for your feedback and contribution!! 🙌

JoshOrndorff:
Needing to use `eprintln` is one more obstacle to frustrate learners. If you don't want to fix it, that's your call.

Each little annoyance that you don't fix (especially when I made the PR for you) is one more reason to find a better tool.

I used polkadot launch successfully in the past, and the build process is much simpler there, so I can actually test my changes and iterate quickly.

pepoviola:
> Needing to use `eprintln` is one more unintuitive workround to frustrate learners.
> 
> If you don't want this PR, that's your call. Each little annoyance that you don't fix is a more reason to find/make/hope for a better tool.

Hey @JoshOrndorff, I understand what you say about `eprintln` and I'm sorry that you find this annoyance. Let me re-work a little bit your pr to see how much overhead we need to support this in both cumulus and non-cumulus based collators. I opened this issue to track https://github.com/paritytech/zombienet/issues/1520. Again, thanks for your feedback and to help us making a better tool.

Thx!

JoshOrndorff:
Actually, this is even more important than I realized. There is a bug in Substrate Node Template for a while now where it occasionally prints `Essential Task Failed` when generating the specs or genesis data. The error is "harmless" in the sense that it doesn't mess up the generated output data. Unless you use the shell redirect.

Although I guess you could argue as well that the `Essential Task Failed` should be logged to stderr. I actually would agree with that point.

But still, this seems like a footgun that is worth avoiding.

And also, just for the record, my collator is exactly the stock cumulus collator. The thing that is unusual about it is that it doesn't use FRAME.

pepoviola:
Hi @JoshOrndorff, this is already fixed and included in the latest version `v1.3.82`.
Thx!

PR #1517, opened 2023-11-16
fix(orchestrator): Apply the x-infra-instance to fileserver pod

### LINKED ISSUE #1513: fileserver pod has no label for scheduling
After the label we agreed to take decision to schedule them on which nodes, it is turned out that fileserver pod doesn't have that label that I would be much appreciated if we can handle that also please.

pepoviola:
- Fix fileserver label for x-infra.
fix #1513

### DIFF
--- a/javascript/packages/orchestrator/src/providers/k8s/kubeClient.ts
+++ b/javascript/packages/orchestrator/src/providers/k8s/kubeClient.ts
@@ -545,7 +545,6 @@ export class KubeClient extends Client {
         type: "deployment",
         files: [
           settings.backchannel ? "backchannel-pod.yaml" : null,
-          "fileserver-pod.yaml",
         ],
       },
     ];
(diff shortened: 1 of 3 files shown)


javascript/packages/orchestrator/src/providers/k8s/kubeClient.ts
```diff
@@ -543,10 +543,7 @@ export class KubeClient extends Client {
       },
       {
         type: "deployment",
-        files: [
-          settings.backchannel ? "backchannel-pod.yaml" : null,
-          "fileserver-pod.yaml",
-        ],
+        files: [settings.backchannel ? "backchannel-pod.yaml" : null],
```
wirednkod, line 546:
```suggestion
        files: settings.backchannel ? ["backchannel-pod.yaml"] : [],
```

wirednkod (approved):
Some nip tuck

javascript/packages/orchestrator/src/providers/k8s/kubeClient.ts
pepoviola: Thanks, I didn't saw this before merge. Can you add it in a follow up pr?
Thx!

PR #1522, opened 2023-11-17
remove shell redirection for cumulus based collators

### LINKED ISSUE #1520: Adjust export commands to work with cumulus/non-cumulus collators
Take this pr as base https://github.com/paritytech/zombienet/pull/1515.

cc: @JoshOrndorff

pepoviola:
Based on https://github.com/paritytech/zombienet/pull/1515, fix #1520 
*NOTE*: we should remove the conditional in the near future.

Also, @wirednkod we should review other places where we are using shell redirections.

cc: @JoshOrndorff

### DIFF
--- a/javascript/packages/orchestrator/src/configGenerator.ts
+++ b/javascript/packages/orchestrator/src/configGenerator.ts
@@ -353,7 +353,13 @@ export async function generateNetworkSpec(
           ? parachain.genesis_state_generator
           : `${collatorBinary} ${DEFAULT_GENESIS_GENERATE_SUBCOMMAND}`;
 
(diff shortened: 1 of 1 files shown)


wirednkod:
As pointed out [here](https://github.com/paritytech/zombienet/pull/1515#issuecomment-1811968348) there are some redirections that need to be added. Should we try to push those in 1 commit?

pepoviola:
> As pointed out [here](https://github.com/paritytech/zombienet/pull/1515#issuecomment-1811968348) there are some redirections that need to be added. Should we try to push those in 1 commit?

I think is better to handle those in a follow up pr, did you want to work on that @wirednkod ?

wirednkod:
Sure thing @pepoviola

pepoviola:
Thanks!!

wirednkod (approved):

Issue #1584, opened 2023-12-05
Nit: Point log links on CI to grafana.teleport.parity.io


skunert:
### Is your feature request related to a problem? Please describe.

On CI, log links are printed and point to `https://grafana.parity-mgmt.parity.io`. 

### Describe the solution you'd like

By pointing to `https://grafana.teleport.parity.io` we could access the logs via google login without having to change the url manually or log into VPN.

### Describe alternatives you've considered

-

PR #1585, opened 2023-12-05
Use grafana teleport URL

skunert:
Closes https://github.com/paritytech/zombienet/issues/1584

Allows usage of grafana logs without vpn.


pepoviola:
Awesome, thanks for your contribution @skunert 🙌

wirednkod (approved):

pepoviola:
I'm not sure why the `nix hash update` check fails since this pr doesn't touch any dep. Could be that the nix config is not reading from the `lock` file?

wirednkod:
> I'm not sure why the `nix hash update` check fails since this pr doesn't touch any dep. Could be that the nix config is not reading from the `lock` file?

https://github.com/paritytech/zombienet/actions/runs/7102648724/job/19333502741?pr=1586

wirednkod:
> > > I'm not sure why the `nix hash update` check fails since this pr doesn't touch any dep. Could be that the nix config is not reading from the `lock` file?
> > 
> > 
> > [paritytech/zombienet/actions/runs/7102648724/job/19333502741?pr=1586](https://github.com/paritytech/zombienet/actions/runs/7102648724/job/19333502741?pr=1586)
> 
> Thanks @wirednkod, I merged :) But I want to understand how to automate this. I think could be because the `orchestrator` is itself a dep of the `cli` but I'm not sure. I will try to check how this works internally. Thx!

Based on a comment, this hash changes with any update of the dependencies, but it is not automatically updated from SRI of package.lock

pepoviola:
> > > > I'm not sure why the `nix hash update` check fails since this pr doesn't touch any dep. Could be that the nix config is not reading from the `lock` file?
> > > 
> > > 
> > > [paritytech/zombienet/actions/runs/7102648724/job/19333502741?pr=1586](https://github.com/paritytech/zombienet/actions/runs/7102648724/job/19333502741?pr=1586)
> > 
> > 
> > Thanks @wirednkod, I merged :) But I want to understand how to automate this. I think could be because the `orchestrator` is itself a dep of the `cli` but I'm not sure. I will try to check how this works internally. Thx!
> 
> Based on a comment, this hash changes with any update of the dependencies, but it is not automatically updated from SRI of package.lock

Yes, and as we are using the `orchestrator` as dep of `cli`  I think any change needs to update the hash. We should automate this in a way that is easy to do it locally.
Thx!

PR #1615, opened 2023-12-22
fix(test-runner): add import into try/catch and show error and stack …

### LINKED ISSUE #1608: Report javascript parse errors
I was running a script:

```zndsl
alice: js-script ./0004-configure-relay.js with "" return is 0 within 600 secs
```

with some code like this:

```js
const calls = [
  api.tx.configuration.setCoretimeCores({ new: 2 }),
  api.tx.coretime.assignCore(0, 20, [[ { task: 100 }, 57600 ]], null)
  api.tx.coretime.assignCore(1, 20, [[ { task: 1005 }, 57600 ]], null)
];
```

and I was getting this error (nothing else):

```
│ 12/19/2023, 12:20:52 PM      │  ❌ alice: js-script ./0004-configure-relay.js with "" return is 0 within 600 secs (37ms)           │
```

Can you see the problem? It took me hours to notice it. I thought it was a problem with `assignCore`!

We should report back if there is a syntax error in the script, instead of failing silently.

pepoviola:
…on debug

fix #1608 

cc: @mrcnski

### DIFF
--- a/javascript/packages/orchestrator/src/test-runner/assertions.ts
+++ b/javascript/packages/orchestrator/src/test-runner/assertions.ts
@@ -324,7 +324,7 @@ const CustomJs = ({
     } catch (err: any) {
       console.log(
         `\n ${decorators.red(
-          `Error importing/running script: ${file_path!}`,
+          `Error running script: ${file_path!}`,
         )} \t ${decorators.bright(err.message)}\n`,
       );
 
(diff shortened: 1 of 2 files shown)


javascript/packages/orchestrator/src/test-runner/assertions.ts
```diff
@@ -320,9 +324,11 @@ const CustomJs = ({
     } catch (err: any) {
       console.log(
         `\n ${decorators.red(
-          `Error running script: ${file_path!}`,
+          `Error importing/running script: ${file_path!}`,
```
mrcnski, line 327: Probably can leave this line unchanged:

```suggestion
          `Error running script: ${file_path!}`,
```

mrcnski (approved):
Great change, thank you!

javascript/packages/orchestrator/src/test-runner/assertions.ts
pepoviola: Yes, from user perspective is the same :)

PR #1638, opened 2024-01-02
fix(orchestrator): use tmp dir when we create paras artifacts

### LINKED ISSUE #1519: Explicitly specify a data directory when exporting parachain genesis state
When zombienet exports the parachain genesis state it uses a command like the following:
```sh
parachain-template-node export-genesis-state --chain /tmp/zombie-91732cca35cc89d9bc40d256bff11459_-279173-rjyW4YgPKWMA/2000-rococo-local.json > /tmp/zombie-91732cca35cc89d9bc40d256bff11459_-279173-rjyW4YgPKWMA/cfg/genesis-state-2000
```

This has worked reliably so far, but only because of a bug in cumulus that I am now fixing. (See https://github.com/paritytech/polkadot-sdk/issues/2326 and https://github.com/paritytech/polkadot-sdk/pull/2331 for context.) Previously, Cumulus made some assumptions about the genesis block, and created a new one on the fly whenever you asked to export the genesis state. With this bug fixed, cumulus will now read the genesis block from the database instead of re-generating it :rocket: 

But that means that you will now need to make sure you are not implicitly using some pre-existing un-purged data directory.

pepoviola:
Use `tmp` dir for `export-genesis-state` cmd.
Fix #1519 
cc: @JoshOrndorff

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -8,7 +8,7 @@
     ...
   }: let
     # this change on each change of dependencies, unfortunately this hash not yet automatically updated from SRI of package.lock
-    npmDepsHash = "sha256-PK1YPfUkLWwzpAQ00S7j5VbfLA8G+zxHaDN4Bjwz7lc=";
+    npmDepsHash = "sha256-RF9jRQFuUNSAkNmpJpiXEpAy5IWBMbd64FVLE4IOavM=";
     ####
 
     # there is officia polkadot on nixpkgs, but it has no local rococo wasm to run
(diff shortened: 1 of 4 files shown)


JoshOrndorff:
The core change looks good to me. Just out of curiosity, why was downgrading chai necessary?

pepoviola:
> The core change looks good to me. Just out of curiosity, why was downgrading chai necessary?

Thanks for the feedback @JoshOrndorff. The issue with `chai` was caused by the depedabot that bump the major version of chai and that version drop `cjs` support, so we need to adjust some parts of our code. I will make the change in a separate pr.

Thx!

pepoviola:
`ci` cache here is causing troubles, I will close this one and re-open with a diff branch.
Thx!

PR #1655, opened 2024-01-08
fix Dockerfile

emamihe:
fix libssl.so.3 issue in k8s-cleaner binary


crates/k8s-cleaner/Dockerfile
```diff
@@ -3,7 +3,9 @@ WORKDIR /usr/src/k8s-cleaner
 COPY . .
 RUN cargo build --release
 
-FROM debian:buster-slim
+#FROM debian:buster-slim
```
pepoviola, line 6:
```suggestion
```

pepoviola (approved):

PR #1677, opened 2024-01-14
Fix #1659, replace whitespace with dash.

### LINKED ISSUE #1659: Zombienet won't allow whitespaces in collator names
SE [issue](https://substrate.stackexchange.com/questions/10776) mentions that names should be passed in quotes like `--name "name"` instead of `--name name`.  
Using zombienet v1.89 i did not verify myself, just opening this so we can close the other.

pepoviola:
fix #1659


wirednkod (approved):

l0r1s:
@pepoviola just an after thought but maybe we could log to the user that we modified a name if we modified it ? wdyt ?

pepoviola:
> @pepoviola just an after thought but maybe we could log to the user that we modified a name if we modified it ? wdyt ?

Yes, sounds good. I think a warning message will be ok. Can you open an issue to track this?
Thx!

PR #1709, opened 2024-01-30
Add check for node version

forgetso:
I've spent far too long today trying to debug "waiting for metrics" errors that were due to fetch not being available in node 16. This PR should prevent other people from making the same mistake.


l0r1s:
Hi @forgetso thanks for your PR, that's true we haven't implemented node version verification and it will ensure this kind of error you encountered.

l0r1s (approved):

PR #1711, opened 2024-01-30
Fixes `clearAuthorities` in chainspec when staking genesis is defined

gpestana:
This PR fixes the `clearAuthorities` operation in chainspec when staking genesis is defined. I'm not 100% sure if the fix is correct, but at least is does not break when the staking pallet has some genesis configs defined as before.

### DIFF
--- a/javascript/packages/orchestrator/src/chainSpec.ts
+++ b/javascript/packages/orchestrator/src/chainSpec.ts
@@ -74,9 +74,13 @@ export function clearAuthorities(specPath: string) {
     if (runtimeConfig?.collatorSelection)
       runtimeConfig.collatorSelection.invulnerables = [];
 
-    // Clear staking
+    // clear staking
     if (runtimeConfig?.staking) {
-      stakingBond = BigInt(runtimeConfig.staking.stakers[0][2]);
+      stakingBond = BigInt(0);
+      for (const staker in runtimeConfig.staking.stakers) {
+        stakingBond += BigInt(runtimeConfig.staking.stakers[staker][2]);
+      }
+
       runtimeConfig.staking.stakers = [];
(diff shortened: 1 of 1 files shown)


javascript/packages/orchestrator/src/chainSpec.ts
```diff
@@ -74,9 +74,13 @@ export function clearAuthorities(specPath: string) {
     if (runtimeConfig?.collatorSelection)
       runtimeConfig.collatorSelection.invulnerables = [];
 
-    // Clear staking
+    // clear staking
     if (runtimeConfig?.staking) {
-      stakingBond = BigInt(runtimeConfig.staking.stakers[0][2]);
```
gpestana, line 79: if the staking genesis is set but with an empty vec of stakers (which is totally OK), accessing `stakers[0][2]` would panic.

wirednkod (approved):

javascript/packages/orchestrator/src/chainSpec.ts
```diff
@@ -74,9 +74,13 @@ export function clearAuthorities(specPath: string) {
...
+      stakingBond = BigInt(0);
+      for (const staker in runtimeConfig.staking.stakers) {
+        stakingBond += BigInt(runtimeConfig.staking.stakers[staker][2]);
+      }
+
```
pepoviola, line 83:
```suggestion
      if ( runtimeConfig.staking.stakers[0] ) stakingBond = BigInt(runtimeConfig.staking.stakers[0][2]);
```
Hi @gpestana, in the current code we only use the value of the first staker for adding the nodes later (without add them). Something like this should work and we can change the first if condition to make the code more clean.
Thx!

PR #1755, opened 2024-03-28
chore: push spawning time metric

pepoviola:
@emamihe this will push the elapsed time to spawn the network with a few labels to query `by_name` or identify the job_id.

Thx!


pepoviola:
Yes, is not deployed yet.
Thx!

PR #1765, opened 2024-04-09
FEAT: updated the keys for scheduling

emamihe:
this PR adds two more options for `x-infra-instance` as:
- ondemand-iops
- spot-iops


pepoviola (approved):
Looks good, thanks!

PR #1774, opened 2024-04-17
added rust installation and PATH manipulation

emamihe:
@pepoviola RUST installation added to the Dockerfile


pepoviola (approved):
thanks!!

Issue #1829, opened 2024-07-11
Support for private kubernetes image registry (imagePullSecrets)


th7nder:
### Is your feature request related to a problem? Please describe.

We're developing a new parachain and wanted to create a private beta with docker image of the parachain published to our ghcr registry. However, we cannot see an option to specify credentials for the registry.

### Describe the solution you'd like

A way to specify [`imagePullSecrets`](https://github.com/paritytech/zombienet/issues?q=imagePullSecrets) to pass to the Kubernetes pod configuration.

pepoviola:
Hi @th7nder, sorry about the delay. We add support for this in the latest version. Please re-open if needed.
Thanks!

PR #1830, opened 2024-07-11
fix: launching from local images in kubernetes

th7nder:
When launching using Kubernetes provider (`zombienet -p kubernetes`) and using minikube, I needed to use local image without publishing it.
I can add it to `minikube`, by using `minikube image load image-name:tag` and specify in `zombienet.toml`:
```toml
[settings]
image_pull_policy = "IfNotPresent"
```

However, it'd still fail at the creating bootnodes (and running --help commands), because the policy was to Always download images and they were not published.

Ideally, this option probably should be sourced from `networkSpec.settings.image_pull_policy`, but I left it at this level.

Let me know what you think!

Sort of related to #1829 (as a workaround to not having private registries)

### DIFF
--- a/javascript/packages/orchestrator/src/providers/k8s/resources/nodeResource.ts
+++ b/javascript/packages/orchestrator/src/providers/k8s/resources/nodeResource.ts
@@ -156,7 +156,7 @@ export class NodeResource {
       {
         image,
         name,
-        imagePullPolicy: "Always",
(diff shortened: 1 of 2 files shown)


pepoviola:
Hi @th7nder, thanks for the feedback. Yes, those pods should read the config from the settings in order to behave like the rest. I can make those changes to fix this.

Thanks!

PR #1842, opened 2024-08-08
Update Install Instructions for NPM and MacOS

shawntabrizi:
The obvious way for me to install zombenet is through some command like `npm`, not to download an try to get a local binary to execute.

Also added more specific instructions for using the binaries on MacOS


pepoviola (approved):
Thanks!!

PR #1860, opened 2024-08-27
bash: use less opionated shebang

FlorianFranzen:
Not every environment has bash installed in `/bin/bash`, so this uses the `env` entrypoint in `/usr/bin` instead to make the bash scripts more portable, e.g. to work on NixOS.


pepoviola (approved):
thanks!

pepoviola (approved):

PR #1862, opened 2024-08-27
readme: mention need for outdate podman

FlorianFranzen:
Let please not waste peoples time by being honest about what is and what is not supported in the Readme!


pepoviola (approved):
Thanks for the feedback, at some point we loose track of the issue. Sorry anout that!

PR #1907, opened 2024-10-15
set 1G as default mem

pepoviola:
cc @emamihe / @skunert


skunert:
Maybe 1 cpu also okay? Or you think thats too much 😬

pepoviola:
> Maybe 1 cpu also okay? Or you think thats too much 😬

We use vms with 8vcpu / 16gb mem. If we set 1cpu as default request I think we will end with less pods by vm and not consuming all the memory. This is only set the default request, the pods can use more if they need (cpu for example). I think this could help but we can also monitoring if pods are not having enough cpu. wdyt?

PR #1940, opened 2025-01-21
Set balance to the double of stakingBond

### LINKED ISSUE #1939: Inc. balance for staking accounts
After https://github.com/paritytech/polkadot-sdk/pull/5501 we should leave some balance (ED) in the staking accounts. (see failure https://gitlab.parity.io/parity/mirrors/polkadot-sdk/-/jobs/8061354).

This could be fixed by double the balance of the staking accounts.
Thx!

pepoviola:
fix #1939 

cc @Ank4n

### DIFF
--- a/javascript/packages/orchestrator/src/chainSpec.ts
+++ b/javascript/packages/orchestrator/src/chainSpec.ts
@@ -135,9 +135,7 @@ export async function addBalances(specPath: string, nodes: Node[]) {
 
         balanceMap[stashKey] = balanceToAdd;
 
-        const logLine = `👤 Added Balance ${
-          balanceToAdd
-        } for ${decorators.green(node.name)} - ${decorators.magenta(stashKey)}`;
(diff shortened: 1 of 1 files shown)


javascript/packages/orchestrator/src/chainSpec.ts
```diff
@@ -130,14 +130,12 @@ export async function addBalances(specPath: string, nodes: Node[]) {
         const balanceToAdd = stakingBond
           ? node.validator && node.balance > stakingBond
             ? node.balance
-            : stakingBond! + BigInt(1)
+            : stakingBond! * BigInt(2) // Double the balance we use for stake
```
Ank4n, line 133: we only need to add ED to staking bond, but this is also fine since staking bond should always be greater than ED.

Ank4n (approved):

javascript/packages/orchestrator/src/chainSpec.ts
pepoviola: Great. Yes, set to the double is an easy fix.

PR #1957, opened 2025-02-11
Fix moonbeam chain spec customizations

RomarQ:
The moonbeam customizations were not correct when using `moonbeam-local` chain specification. The reason is that `moonbeam` uses a [SUPPLY_FACTOR](https://github.com/moonbeam-foundation/moonbeam/blob/4c9dc6d12d82108987d15d29f49f29af034e2398/runtime/moonbeam/src/lib.rs#L138) of `100`, while `moonbase` and `moonriver` have a supply factor of `1`. This was causing the `reservedBalance` to be incorrect.

The fix included in this PR, increases `stakingBond` and `reservedBalance` default values to a reasonable amount and overrides the balance of the collator if it is not enough to cover the staking bond and the authoring deposit.

### DIFF
--- a/javascript/packages/orchestrator/src/chain-decorators/moonbeam.ts
+++ b/javascript/packages/orchestrator/src/chain-decorators/moonbeam.ts
@@ -141,7 +141,9 @@ async function addParaCustom(specPath: string, node: Node) {
 
   const expectedBalance = stakingBond + reservedBalance;
   if (collatorBalance) {
-    collatorBalance[1] = expectedBalance;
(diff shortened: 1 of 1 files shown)


pepoviola (approved):
Looks good to me, thanks!!

RomarQ:
Hi @pepoviola,

Do you plan to do a new zombienet release this week?

pepoviola:
> Hi @pepoviola,
> 
> Do you plan to do a new zombienet release this week?

Yes, I will release a new version today.
Thanks!!

PR #1958, opened 2025-02-12
ci: use gcp registry for integration tests

alvicsam:
PR changes integration tests workflow so it pushes images to GCP registry. I also decided to keep docker image build for master but it will push to docker hub.

cc https://github.com/paritytech/ci_cd/issues/1106


pepoviola (approved):

pepoviola:
thanks @alvicsam 🙌

PR #1973, opened 2025-03-03
[native] use `local_ip` settings for p2p in native

pepoviola:
Allow to use `local_ip` as `listen-addr` (p2p) for native provider.


pepoviola:
This needs to be adapted to the current code.

PR #2007, opened 2025-03-27
Dropping bucket metrics in podMonitor

lazam:
Buckets are generated by the polkadot-sdk, but I don't think it is beneficial with zombienet's use case since it is expected to be short-lived. A single bucket metric (e.g, substrate_tasks_polling_duration_bucket) creates around 6090 series for a single zombienet namespace, and there's a lot of metrics aside from substrate_tasks_polling_duration_bucket. Dropping the histogram (bucket) metrics are beneficial since we can still refer to other metrics (e.g, _sum, _count, _total).

### DIFF
--- a/javascript/packages/orchestrator/static-configs/pod-monitor.yaml
+++ b/javascript/packages/orchestrator/static-configs/pod-monitor.yaml
@@ -11,7 +11,7 @@ spec:
       port: prometheus
       metricRelabelings:
         - sourceLabels: [__name__]
-          regex: ".*_bucket"
+          regex: '^(?!polkadot_pvf_.*_bucket$).+_bucket$'
           action: drop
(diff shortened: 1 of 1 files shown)


lazam:
> polkadot_pvf



> Hi @lazam, thanks for checking this but `zombienet` allow to query histograms and we are using this two in our current pipeline
> 
> ```
> polkadot_pvf_execution_time
> polkadot_pvf_preparation_time
> ```
> 
> Can we keep those two?
> 
> Thx!

Oh I didn't know about that, I excluded metrics with `polkadot_pvf` now since those metrics are minimal compared to others so we don't need to add them in the near future if needed.

pepoviola (approved):
thanks!!!

PR #2024, opened 2025-06-17
Fixed regression while binding localhost on native

sigurpol:
Fixed the regression introduced by [#2020](https://github.com/paritytech/zombienet/pull/2020) where the invalid local address `127.0.01` was used.


sigurpol:
@pepoviola  just made and merged #2025  that does exactly the same as this PR so I can close this one. Thanks for the quick fix!!!

Issue #2077, opened 2025-08-26
A possibility to re-use base-path for config arguments


evgeny-s:
### Is your feature request related to a problem? Please describe.

Hey, guys.
Is there a possibility to reuse the auto-generated `--base-path` for the test execution in the other arguments?
The use case is that we are running zombienet node for an embedded node, and we need to specify a bunch of argument groups. 
Would be great to have something like:
```
command: "binary arg1 -- ",
args: [
    '-- --base-path={ZOMBIENET_TMP_DIR}/container-chain-data'
]
```

Before spawning the node, it could parse all the arguments and substitute the auto-generated default execution folder path.

Thanks!

### Describe the solution you'd like

The solution mentioned above

pepoviola:
Hey @evgeny-s, thanks for the feedback! Yes, I think we can add the `base_path` as part of the token replacements. Let me review it and ping you back.
Thx!

cc: @RafalMirowski1

PR #2094, opened 2025-09-17
Reuse base-path for config arguments

DenzelPenzel:
#2077 
- Added `.idea` to `.gitignore` to exclude IDE-specific files.
- Introduced a new example configuration file `0006-reuse-base-path.toml` for relaychain and parachains

### DIFF
--- a/examples/0006-reuse-base-path.toml
+++ b/examples/0006-reuse-base-path.toml
@@ -1,3 +1,15 @@
+# Network Token Replacement Example: Custom Base Path
+#
+# This example demonstrates how to use network token replacement to customize
+# the --base-path argument while reusing zombienet's auto-generated base directory.
+#
+# The token {{'network'|zombie('base_path')}} gets replaced with the network's
+# temporary directory path at runtime (e.g., /tmp/zombie-abc123/alice/data),
+# allowing you to append custom suffixes like "/alice-custom-data".
+#
+# Available network token:
+# - {{'network'|zombie('base_path')}}  - Node-specific data path (tmpDir + /custom_data_path)
+
 [relaychain]
 default_image = "docker.io/paritypr/polkadot-debug:master"
 default_command = "polkadot"


pepoviola:
Hey @DenzelPenzel, I think is better to re-use the custom token machinery we already have here

https://github.com/paritytech/zombienet/blob/727c2c4c916cd874067589e58157f445a1f93067/javascript/packages/orchestrator/src/network.ts#L384-L392

but that machinery is tied to the pattern `{{ZOMBIE:node_name:key}}` so maybe we can extend to also include the `network` instance properties and will be something like `{{ZOMBIE:<node_name|network>:key` and we make clear that __network__ is a reserved word for this.

So, from the config the could do something like
{{'network'|zombie('base_path')}} -> that internally we translate to `ZOMBIE:network:base_path` and then should be replaced with networkInfo.

make sense? if you are agree we can touch base tomorrow on this.

Thx!

DenzelPenzel:
> but that machinery is tied to the pattern `{{ZOMBIE:node_name:key}}` so maybe we can extend to also include the `network` instance properties and will be something like `{{ZOMBIE:<node_name|network>:key` and we make clear that **network** is a reserved word for this.
> 
> So, from the config the could do something like {{'network'|zombie('base_path')}} -> that internally we translate to `ZOMBIE:network:base_path` and then should be replaced with networkInfo.



> Hey @DenzelPenzel, I think is better to re-use the custom token machinery we already have here
> 
> https://github.com/paritytech/zombienet/blob/727c2c4c916cd874067589e58157f445a1f93067/javascript/packages/orchestrator/src/network.ts#L384-L392
> 
> but that machinery is tied to the pattern `{{ZOMBIE:node_name:key}}` so maybe we can extend to also include the `network` instance properties and will be something like `{{ZOMBIE:<node_name|network>:key` and we make clear that **network** is a reserved word for this.
> 
> So, from the config the could do something like {{'network'|zombie('base_path')}} -> that internally we translate to `ZOMBIE:network:base_path` and then should be replaced with networkInfo.
> 
> make sense? if you are agree we can touch base tomorrow on this.
> 
> Thx!

sounds good, let me check it

examples/0006-reuse-base-path.toml
```diff
@@ -0,0 +1,25 @@
+[relaychain]
```
pepoviola, line 1: Maybe we can add an small docs on top of this network config describing the example (e.g in this case the use of the `base_path` )
Thx!

pepoviola (approved):
Small nits about docs.

PR #2100, opened 2025-09-29
Add flag exclusion support with -:<flag> syntax

### LINKED ISSUE #1737: An option to turn off `--insecure-validator-i-know-what-i-do`
### Is your feature request related to a problem? Please describe.

By default in `validator` mode `--insecure-validator-i-know-what-i-do` flag is passed by [default](https://github.com/paritytech/zombienet/blob/main/javascript/packages/orchestrator/src/cmdGenerator.ts#L273). But as far as I know there is no way to turn it off - this flag doesn't support taking a value. So I couldn't test a secure validator mode (or at least I haven't found away and this should be documented somewhere).

Related issue https://github.com/paritytech/zombienet/issues/1551)

### Describe the solution you'd like

I would like to have a switch to exclude the flag from the default args

### Describe alternatives you've considered

_No response_

### Additional context

_No response_

DenzelPenzel:
fix #1737


docs/src/network-definition-spec.md
```diff
@@ -43,7 +43,8 @@ The network config can be provided both in `json` or `toml` format and each sect
   - `image`: (String) Override default docker image to use for this node.
   - `command`: (String) Override default command.
   - `command_with_args`: (String) Override default command and args.
-  - `args`: (Array of strings) Arguments to be passed to the `command`.
+  - `args`: (Array of strings) Arguments to be passed to the `command`. 
+    - **Flag Exclusion**: You can exclude default flags by prefixing them with `-:`. For example, `-:--insecure-validator-i-know-what-i-do` will exclude the default `--insecure-validator-i-know-what-i-do` flag that is normally added to validator nodes. Both full flag names (e.g., `-:--flag-name`) and short names (e.g., `-:flag-name`) are supported.
```
pepoviola, line 47: Maybe is a good idea to also list all the __default__ flags we add.

pepoviola (approved):
small _nit_ about docs and the is good to merge.

PR #2158, opened 2026-04-06
nix: use nixpkgs-unstable

andresilva:
Switch `nixpkgs` input from `nixos-23.11` (severely outdated!) to `nixpkgs-unstable`: this flake is not a NixOS system configuration, so tracking a stale NixOS release branch unnecessarily limits package freshness, `nixpkgs-unstable` is the standard moving target for package use and remains pinned via `flake.lock` for reproducibility, and downstream users can override the `nixpkgs` input if they need a different pin (to use a stable release for example). Also fixes the issue with latest `nixpkgs` where `nodePackages` is gone.

Fixes https://github.com/andresilva/polkadot.nix/actions/runs/24026843959/job/70067108329.

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -233,8 +233,8 @@
           nativeBuildInputs = with pkgs;
             [
               python3
-              nodePackages.node-gyp-build
-              nodePackages.node-gyp
+              node-gyp-build
+              node-gyp
             ]
             ++ pkgs.lib.optional pkgs.stdenv.isDarwin (with pkgs;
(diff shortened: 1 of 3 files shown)


pepoviola:
> I'm down to package that and distribute in polkadot.nix (to replace existing zombienet) once it's stable/released.

The new version is already stable, and available in both crates.io and gh releases. Thanks!!

andresilva:
I have packaged zombienet-sdk now: https://github.com/andresilva/polkadot.nix/pull/95

PR #2180, opened 2026-06-22
`(default_)db_snapshot` handles 'file:://' paths

andrzejSulkowski:
We are about to setup zombienet for the Individuality repo which includes a bunch of long running initialization scripts. To not require to init the network on every run, we intend to let a GitHub runner yield periodically a initialized db as an artifact which can be download and re-used.

However, the only accepted source is `http(s)://`. We end up downloading the db, then re-running a http server which hosts the db so zombienet can download it into its directory.

**Solution:**
Accept `file://` paths.


pepoviola:
Hey @andrzejSulkowski, thanks for your contribution. We are trying to deprecate this version in favor of the new one [zombienet-sdk](https://github.com/paritytech/zombienet-sdk) that allow to use an `url` or a `file` path as db_snapshot. Can you use the new version to setup the tests?

Thx!

andrzejSulkowski:
closed in favor of using [zombienet-sdk](https://github.com/paritytech/zombienet-sdk)

